Issue #742, opened 2023-06-02
How to make a package with components that embed components?


jrray:
I'm getting tripped up on trying to make a package that has components where those components are expected to embed specific components of some other package. In the case of boost and boost-python, here's a simplified concrete example of what I'm trying to do:

boost:build contains header.hpp
boost:run contains libwhatever.so.1.2.3, and the "build" component has `uses: [run]`.

The boost-python package is a full build of boost with python stuff enabled at build time, so it's split up into components in a way that mirrors the boost package. But the boost-python "build" component should embed the boost "build" component, likewise the "run" component should embed the boost "run" component.

What I've tried looks like this (boost-python.spk.yaml):
```yaml
install:
  components:
    - name: build
      files:
        - include/
        - lib/cmake/
        - lib/*.a
        - lib/*.so
      uses:
        - run
      embedded:
        - pkg: boost/1.76.0+r.3
          install:
            components:
              - name: build
                uses:
                  - run
    - name: run
      files:
        - "*"
      file_match_mode: Remaining
      embedded:
        - pkg: boost/1.76.0+r.3
          install:
            components:
              - name: run
```

I'm not sure about the `embedded` entries here and I don't believe this is accomplishing what I'm looking for. For example, I made this extra dummy package:
```yaml
pkg: dummy/1.0.0

sources: []

build:
  script: []

install:
  requirements:
    - pkg: boost:build/1.76.0+r.3
```

Note that it specifically depends on the "build" component of boost.

Then, I do this contrived thing: `spk env -vv dummy boost-python:run`. I ask for the dummy package and boost-python:run.

It solves and I get an environment like this:
```
Installed Packages:
  local/dummy:{build,run}/=1.0.0/3I42H3S6
  local/boost-python:run/=1.76.0+r.3/Q4VLNGUE
  boost/=1.76.0+r.3/embedded
```

What's wrong here is that boost:build is not satisfied. This environment doesn't actually have any of the boost[-python] build component's files in it. This environment needs to include `boost-python:build` to actually satisfy dummy's requirements. The presence of `boost/=1.76.0+r.3/embedded` appears to satisfy the solver but the environment is missing required files.

PR #744, opened 2023-06-02
Rework how packages describe how they contain the components of an embedded package

jrray:
Addresses the question from #742.

The original spec schema had multiple places where an embedded package could be defined:

```yaml
pkg: demo/1.0.0
install:
  components:
    - name: comp1
      embedded:
        # inside a component
        - pkg: embedded/2.0.0
  embedded:
    # outside a component
    - pkg: embedded/2.0.0
```

It becomes confusing if the embedded package also has components. Where do those components get defined? How can the parent package say which of its components contains the embedded packages' components?

## A motivating example

The package boost contains a lot of headers only needed during build so it benefits from having those files live only in the "build" component. A second package, boost-python, is also a full build of boost (plus extra python things), so it needs to embed boost. Since the boost-python package is a full build of boost, it also benefits from having the headers only live in the "build" component. It needs to be able to say that the `boost-python:build` component embeds the `boost:build` component.

```yaml
pkg: boost-python/1.76.0
install:
  components:
    - name: build
      files:
        - include/
        - lib/cmake/
        - lib/*.a
        - lib/*.so
      uses:
        - run
      embedded_components:    # <-- new
        - boost:build
    - name: run
      files:
        - "*"
      file_match_mode: Remaining
      embedded_components:    # <-- new
        - boost:run
  embedded:
    - pkg: boost/1.76.0
      install:
        components:
          - name: build
            uses:
              - run
          - name: run
```

This shows how the boost-python spec can now be written. The embedded boost package is defined once under `install.embedded` and its components are defined here. The new `install.components[].embeded_components` field is used to specify for each component of boost-python, which components of boost are present.

## Solver logic changes

There can be a situation where a certain combination of requests will cause the solver to fail even though there is nothing "wrong" about those requests.

First, there needs to be a request for a specific component of a real package: `real-package:comp1`. Assuming `real-package:comp1` exists, this request will solve successfully.

Then, some other package needs to exist that embeds `real-package`, let's call it `fake-package`. Assume `fake-package:comp1` embeds `real-package:comp1` and other components exist inside `fake-package`.

Now, add a request for a specific component of `fake-package` that isn't `comp1`: `fake-package:comp2`. As in, `spk env real-package:comp1 fake-package:comp2`.

The problem arises when `fake-package` replaces `real-package` in the solve, but only `fake-package:comp2` is part of the solution, because this does not include (embed) `real-package:comp1`.

The solver now needs to detect that the current set of requests do not bring in all the necessary components of the parent package in order to have the required components of the embedded package satisfied. It effectively needs to translate the `real-package:comp1` request into a request for `fake-package:comp1`. When this situation is detected, the state will be rolled back to just before `fake-package` is added to the solution, and a request for `fake-package:comp1` is added. This combines with the explicit request for `fake-package:comp2`, making the merged request into `fake-package:{comp1,comp2}`. Now when `fake-package` is added to the solution it will include `comp1` and the request for `real-package:comp1` will be satisfied.

### DIFF
--- a/crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
+++ b/crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
@@ -6,7 +6,7 @@ use std::collections::BTreeSet;
 
 use nom::character::complete::char;
 use nom::combinator::{cut, map, opt};
-use nom::error::{ContextError, ParseError};
+use nom::error::{ContextError, FromExternalError, ParseError};
 use nom::sequence::{pair, preceded};
 use nom::IResult;
 use nom_supreme::tag::TagError;
@@ -15,6 +15,8 @@ use crate::ident_component::parsing::components;
 use crate::ident_component::Component;
 use crate::name::parsing::package_name;
 use crate::name::PkgName;
+use crate::version::parsing::version;
+use crate::version::Version;
 
 /// Parse a package name in the context of a range identity.
 ///
@@ -37,5 +39,41 @@ where
             opt(preceded(char(':'), cut(components))),
             |opt_components| opt_components.unwrap_or_default(),
         ),
+    )(input)
+}
+
+/// Parse a package name with optional components and optional version.
+///
+/// The package name must either be followed by a `/` or the end of input.
+///
+/// Examples:
+/// - `"package-name"`
+/// - `"package-name/"`
+/// - `"package-name/1.0.0"`
+/// - `"package-name:comp"`
+/// - `"package-name:{comp1,comp2}/"`
+/// - `"package-name:{comp1,comp2}/1.0.0"`
+pub fn request_pkg_name_and_version<'a, E>(
+    input: &'a str,
+) -> IResult<&'a str, (&PkgName, BTreeSet<Component>, Option<Version>), E>
+where
+    E: ParseError<&'a str>
+        + ContextError<&'a str>
+        + FromExternalError<&'a str, std::num::ParseIntError>
+        + FromExternalError<&'a str, crate::version::Error>
+        + TagError<&'a str, &'static str>,
+{
+    map(
+        pair(
+            package_name,
+            pair(
+                map(
+                    opt(preceded(char(':'), cut(components))),
+                    |opt_components| opt_components.unwrap_or_default(),
+                ),
+                opt(preceded(char('/'), cut(version))),
+            ),
+        ),
+        |(pkg_name, (components, opt_version))| (pkg_name, components, opt_version),
     )(input)
 }
--- a/crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
+++ b/crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
@@ -44,7 +44,8 @@ where
 
 /// Parse a package name with optional components and optional version.
 ///
-/// The package name must either be followed by a `/` or the end of input.
+/// The package name must either be followed by a `/` and a version, or the end
+/// of input.
 ///
 /// Examples:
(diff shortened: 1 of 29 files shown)


jrray:
Another angle to this problem came up today. Although `argparse` is bundled in the python standard library, it is also sometimes listed as a dependency of packages in pypi and in our environment the use of `spk convert pip -- unittest2` ended up creating and depending on a `python-argparse` package.

I'd say there's a general problem here with respect to packages bundled in the system library that are also in pypi, and there's a category of that problem that argparse falls into where the package is old and not getting updates.

The `python` package basically needs a way to specify that it contains `python-argparse`, but the twist is that we don't want it to prevent (in the general sense) allowing a newer version of `python-argparse` to be brought into the environment. So part of this requires the functionality in this PR to add `python-argparse` as both a component and embedded, then the expected behavior would be:

1. make the `argparse` component included in the environment by default (because it is expected to be there via the standard library); but
2. allow a request for `python-argparse` to overrule the default request for the `argparse` and conflict with an explicit request for the `argparse` component; and
3. prevent the external `python-argparse` package from being a version lower than the version in the `argparse` component.

I think this might all be possible with this PR and the existing `install.requirements` functionality.

1. the `argparse` component will embed `python-argparse` but not be included by default
2. the top-level python package will have an `install.requirement` for `python-argparse` of at least the version that is bundled in python

When solving for python, the `install.requirement` will force the solver into either picking up a newer "real" `python-argparse` or it will pick up the `argparse` component via finding it through an embedded stub.

What I'm not sure already works [with this PR] is the solver finding a stub for an embedded package that comes from a component and having that add a request for that component to an existing request for the provider package.

rydrman:
From the meeting today:

  ```yaml
  install:
    components:
      - name: cmpt
        embedded: 
          - name:component
    embedded:
      - pkg: name
  ```

jrray:
> From the meeting today:
> 
>     * default `run` components embed all `install.embedded` entries
> 
>     * default `build` components DO NOT, but they do need to `use` the `run` component

I'm not convinced that the extra complexity of implementing this adds any value over what this PR is already doing. If `install.components[].embedded` is empty, it is given a "fabricated" value where the components of `install.embedded[]` are mapped 1:1.

So for example:

```yaml
pkg: demo/1.0.0
install:
  embedded:
    - pkg: embedded1/1.0.0
    - pkg: embedded2/1.0.0
```

expands to:

```yaml
pkg: demo/1.0.0
install:
  components:
    - name: build
      embedded:
        - embedded1:build/1.0.0
        - embedded2:build/1.0.0
    - name: run
      embedded:
        - embedded1:run/1.0.0
        - embedded2:run/1.0.0
  embedded:
    - pkg: embedded1/1.0.0
      install:
        components:
          - name: build
          - name: run
    - pkg: embedded2/1.0.0
      install:
        components:
          - name: build
          - name: run
```

Changing how the default build and run components are generated feels a bit out of scope for this PR.

crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
```diff
@@ -39,3 +41,39 @@ where
         ),
     )(input)
 }
+
+/// Parse a package name with optional components and optional version.
+///
+/// The package name must either be followed by a `/` or the end of input.
```
rydrman, line 47: This line seems like it's wrong for this function...

rydrman (approved):

crates/spk-schema/crates/foundation/src/ident_ops/parsing/request.rs
jrray: It's not wrong but I changed it a little.

PR #851, opened 2023-08-25
Capture unrecognized fields with lint

nicholy:
PR containing the current progress of updating the `Spk lint` command to capture typos and unrecognized config fields within the spk spec file. The current approach taken is to modify how the deserialize method works. 

Example:

```
install:
  envionment:
    - priority: 99
  requirements:
    - pkg: python
      # we can use the version of python from the build environment to dynamically
      # define the install requirement
      fromBuildEnv: x.x
```

Outputs the error:

```
Failed temp-pkg.spk.yaml:
-----> Unrecognized key: enviroment. (Did you mean: 'environment'?)
```

### DIFF
--- a/crates/spk-cli/group4/src/cmd_lint.rs
+++ b/crates/spk-cli/group4/src/cmd_lint.rs
@@ -8,9 +8,8 @@ use clap::Args;
 use colored::Colorize;
 use miette::Result;
 use spk_cli_common::{flags, CommandArgs, Run};
-use spk_schema::v0::Spec;
 use spk_schema::Lint::Key;
-use spk_schema::{AnyIdent, LintedItem, SpecTemplate, Template, TemplateExt};
+use spk_schema::{FromYaml, LintedSpec, SpecTemplate, Template, TemplateExt};
 
 /// Validate spk yaml files
 #[derive(Args)]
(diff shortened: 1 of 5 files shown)


crates/spk-cli/group4/src/cmd_lint.rs
```diff
@@ -25,12 +27,26 @@ impl Run for Lint {
...
     async fn run(&mut self) -> Result<Self::Output> {
-        let options = self.options.get_options()?;
         let mut out = 0;
+        let options = self.options.get_options()?;
         for spec in self.packages.iter() {
-            let result = SpecTemplate::from_file(spec).and_then(|t| t.render(&options));
-            match result {
-                Ok(_) => println!("{} {}", "OK".green(), spec.display()),
+            let yaml = SpecTemplate::from_file(spec).and_then(|t| t.render_to_string(&options))?;
+            let lints: std::result::Result<LintedItem<Spec<AnyIdent>>, serde_yaml::Error> =
```
jrray, line 34: I think it is a problem if this is hard coded to expect `spk_schema::v0::Spec` here, rather than the more general `spk_schema::Spec`. The lint command needs to work for all spec types/versions.

crates/spk-schema/src/v0/spec.rs
```diff
@@ -990,51 +1122,46 @@ where
...
-                "tests" => self.tests = Some(map.next_value::<Vec<TestSpec>>()?),
-                "install" => self.install = Some(map.next_value::<InstallSpec>()?),
-                _ => {
-                    // ignore any unrecognized field, but consume the value anyway
-                    // TODO: could we warn about fields that look like typos?
+                "sources" => self.sources = Some(map.next_value::<Vec<LintedItem<SourceSpec>>>()?),
+                "build" => {
+                    self.build = {
+                        let build = map.next_value::<LintedItem<UncheckedBuildSpec>>()?;
+                        let lints = build.lints.clone();
```
jrray, line 1132: It doesn't look a clone would be necessary here. Maybe a `take` instead?

Issue #923, opened 2023-11-28
`spfs clean` fails if disk is full


jrray:
When the local spfs repository is located on a disk that is very full, `spfs clean` will fail with "No space left on device" if the user running the command does not already have a renders directory. As the only way to open a local repository is to ultimately `create` it, this involves attempting to create a number of directories, including the per-user renders directory.

This has to be worked around by either running the clean as a different user that already has a renders directory, or manually removing some files to free up enough space to let clean run normally.

The `clean` command has no need for that renders directory to exist, but there is currently no way for `clean` to just open an existing repo in a "read only" way.

jrray:
#1159 will address this issue.

PR #924, opened 2023-11-30
Draft: explore type state for FSRepository

jrray:
Exploring the idea of using the type state pattern to have two different flavors of an `FSRepository`, one that promises the user's renders directory exists, and one that doesn't, for use by cmd_clean (#923).

There are some type conversion todos left in the code, but I got this to the point where the non-test code will compile and in theory `spfs clean` would not attempt to create the user's render directory.

Looking for feedback on this approach or if there are any suggestions for an alternative approach. Something much simpler would be the tried and true pass-a-bool-down, since all I really want to accomplish is to skip creating one of the directories inside `create`. There's just a lot of abstraction between `open_repository_from_string` and the call to `create`.

This draft change doesn't offer any other advantages, like it doesn't prevent using a no-renders repo as if it were a yes-renders one. The distinction gets hidden behind the `RepositoryHandle`. I guess I'm saying I don't like this code but I thought I'd at least get a second opinion.

### DIFF
--- a/crates/spfs-cli/cmd-clean/src/cmd_clean.rs
+++ b/crates/spfs-cli/cmd-clean/src/cmd_clean.rs
@@ -8,6 +8,7 @@ use chrono::prelude::*;
 use clap::Parser;
 use colored::Colorize;
 use miette::Result;
+use spfs::storage::fs::NoRenderStoreForCurrentUser;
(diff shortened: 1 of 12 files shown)


jrray:
Superseded by #1159.

PR #1064, opened 2024-07-02
Draft: Add loom simulations of spfs clean data race

jrray:
$ cargo test --bin clean-loom

This shows three variations on how `spfs clean` can race with writers to a repo.

The first one tries to simulate the existing behavior and the test fails as expected.

The other two demonstrate how a write-ahead log might prevent races. The first one requires the write-ahead log to be locked while reading in all the tags, which in turn would block writes for a significant amount of time.

The second one proposes how the write-ahead log could be pruned over time with a kakfa-like tombstone concept, and only requires the log to be locked while actually deleting files.

In either case, the time while the write-ahead log is locked could early exit in order to not block writers for an extended amount of time. In this way it could make incremental progress by deleting some number of objects in each pass.

### DIFF
--- a/clean-loom/src/main.rs
+++ b/clean-loom/src/main.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 
 use loom::sync::{Arc, Mutex};
 use loom::thread;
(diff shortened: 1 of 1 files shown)


jrray:
> The server side would mark as many of these as possible (including children of these objects).

Thinking about this more, since the cleaner is not allowed to create dangling pointers, then for pushing existing objects that can be marked in time, it wouldn't be necessary to mark the children. Keeping the parent alive keeps all the children alive.

PR #1067, opened 2024-07-04
Add `--output-to-dir <path>` for solvers' output

dcookspi:
Adds `--output-to-dir <path>` flag for logging each solver's output to a separate file for each solver run, and the ` SPK_OUTPUT_TO_DIR` env avr. The output files will be put in the given path.

It extends the `OutputKind` enum to have entries that output to a file, or `println!` and a file, or `tracing!` and a file. The interface allows for anything with the `Write` trait (and some thread-safe/task required traits). The `Copy` trait was removed from the enum in favour of cloning while wrangling the tasks, threads, mutability, and task communication messages. Let me know if there's a better way to do this.

A method was added to the enum to allow the initial output kind, i.e. `Println` or `Tracing`, to be combined with an output to file enum entry. This is used to make things appear in the right places when a background solver finishes the solve first and needs to print the solve results for the user (`Installed Packages: ...`).

The solve output files are named with the date and time the solve started, and the name of the kind of solver,  e.g.
- `<path>`/`spk_solver_run_20240704_131234_12004596_Unchanged`
- `<path>`/`spk_solver_run_20240704_131234_23142152_All-Impossible-Checks`

The verbosity level is increased to at least 2 when `--output-to-dir ...` is used. This helps ensure the files contain enough solver information to be useful when debugging issues. This minimum verbosity level is configurable via `--output-to-dir-min-verbosity <n>` or the `SPK.SPK_OUTPUT_TO_DIR_MIN_VERBOSITY` env var).

### DIFF
--- a/crates/spk-solve/src/io.rs
+++ b/crates/spk-solve/src/io.rs
@@ -7,7 +7,7 @@ use std::collections::VecDeque;
 use std::fmt::{Display, Write};
 use std::fs::File;
 use std::io::{BufWriter, Write as IOWrite};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
@@ -502,7 +502,8 @@ pub struct DecisionFormatterBuilder {
     stop_on_block: bool,
     step_on_block: bool,
     step_on_decision: bool,
-    output_to_file: bool,
+    output_to_dir: Option<PathBuf>,
+    output_to_dir_min_verbosity: u8,
 }
 
 impl Default for DecisionFormatterBuilder {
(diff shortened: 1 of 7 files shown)


crates/spk-solve/src/io.rs
```diff
@@ -700,17 +733,128 @@ impl Pluralize for str {
...
+                // valid for combining with the current output kind here.
+                Err(Error::IncludingThisOutputNotSupported("OutputKind::ensure_output must be called with a Println or Tracing other_output. Including other kinds".to_string()))
+            }
+        }
+    }
+
+    /// Flush the output kind's output file, if any. Any errors that
+    /// occur while flushing the file are logged but not returned
+    /// because we don't want them to interrupt the solve.
+    fn flush(&self) {
```
jrray, line 836: I'm not sure I agree with this approach because the reason an explicit flush exists is to surface any errors. But I do recognize this is a private method and logging a warning is likely the only way we want to proceed anyway.

If this was a public API I would want this to return `Result` so the caller can decide how to deal with errors.

jrray (approved):

PR #1086, opened 2024-08-02
Changes needed for using Rust 1.80

jrray:
This does not bump Rust to 1.80 yet but makes changes necessary to compile and pass lints. The `std::env` related changes may need a closer look, please refer to the individual commits.

Edit: I should mention that I did not want to add the `unsafe` keyword since I didn't think it could be justified as safe.

### DIFF
--- a/crates/spfs/src/graph/annotation.rs
+++ b/crates/spfs/src/graph/annotation.rs
@@ -92,18 +92,14 @@ impl<'a> AnnotationValue<'a> {
     }
 
     /// Note: This is used for legacy and new style digest calculations
-    pub fn legacy_encode(&self, mut writer: &mut impl std::io::Write) -> Result<()> {
+    pub fn legacy_encode(&self, writer: &mut impl std::io::Write) -> Result<()> {
         match self {
             AnnotationValue::String(v) => {
-                // Clippy doesn't realize `writer` can't be moved here.
-                #[allow(clippy::needless_borrows_for_generic_args)]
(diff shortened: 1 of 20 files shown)


crates/spfs/src/graph/annotation.rs
```diff
@@ -95,10 +95,14 @@ impl<'a> AnnotationValue<'a> {
     pub fn legacy_encode(&self, mut writer: &mut impl std::io::Write) -> Result<()> {
         match self {
             AnnotationValue::String(v) => {
+                // Clippy doesn't realize `writer` can't be moved here.
+                #[allow(clippy::needless_borrows_for_generic_args)]
                 encoding::write_uint8(&mut writer, AnnotationValueKind::String as u8)?;
```
rydrman, line 100: I did a little digging into this one and it looks like the proper solution to avoid this lint is to re-borrow rather than creating a double-reference... did you try this pattern?

```suggestion
                encoding::write_uint8(&mut *writer, AnnotationValueKind::String as u8)?;
```

jrray: I didn't try that. I'm traveling this weekend so if you want to take over to get this change in and merged feel free.

rydrman: Will do, I've moved the actual CI/builds as well now and will merge if things pass

jrray: Thanks!

PR #1087, opened 2024-08-03
Allow building packages without an origin configured

### LINKED ISSUE #1085: Doc improvement - spfs config remote
**Is your feature request related to a problem? Please describe.**
Following the doc chapter installation and bootstrapping option1 result in:

```bash
spk build python.spk.yaml
 INFO rendering template for python
 INFO saving package recipe for python/3.11.9
 INFO collecting sources for python/3.11.9
 INFO Validating source package contents...
 INFO Committing source package contents...
Error: spfs::unknown_remote (https://spkenv.dev/error_codes#spfs::unknown_remote)

  × No remote named 'origin' configured
  help: See available remotes via the 'spfs config' command

Error:   × Failed to spawn spfs-render
  ╰─▶ process exited with non-zero status

ERROR spfs::generic (https://spkenv.dev/error_codes#spfs::generic)

  × Failed to collect sources
  ╰─▶ Failed to re-mount runtime filesystem: spfs-enter --remount failed with code Some(1)
  ```

**Describe the solution you'd like**
I guess a step is missing in the documentation to setup spfs config remote before building package

### PR DESCRIPTION
**Pull request by rydrman**
Replaces the default use of the origin repo with a check that only does so if it is actually configured. This will allow for building packages locally on a machine without first setting up a configuration file.

closes #1085

### DIFF
--- a/crates/spk-cli/common/src/flags.rs
+++ b/crates/spk-cli/common/src/flags.rs
@@ -20,6 +20,7 @@ use spk_schema::foundation::option_map::OptionMap;
 use spk_schema::foundation::spec_ops::Named;
 use spk_schema::foundation::version::CompatRule;
 use spk_schema::ident::{parse_ident, AnyIdent, PkgRequest, Request, RequestedBy, VarRequest};
+use spk_schema::ident_ops::NormalizedTagStrategy;
 use spk_schema::option_map::HOST_OPTIONS;
 use spk_schema::{Recipe, SpecRecipe, SpecTemplate, Template, TemplateExt, TestStage, VariantExt};
 #[cfg(feature = "statsd")]
@@ -957,7 +958,7 @@ impl Repositories {
         if self.local_repo_only {
             return Ok(repos);
         }
-        for (name, ts) in enabled.into_iter().chain([("origin", None)]) {
+        for (name, ts) in enabled.into_iter() {
             if disabled.contains(name) {
                 continue;
             }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crates/spk-cli/common/src/flags.rs [hunk 1]
```diff
@@ -980,6 +981,21 @@ impl Repositories {
...
+        let has_origin = repos.iter().find(|(n, _)| n == "origin").is_some();
+        if !disabled.contains("origin") && !has_origin {
+            // the origin repo is considered a default and specifically
+            // added at the very end of any command line flags if needed
+            match storage::remote_repository::<_, NormalizedTagStrategy>("origin").await {
+                Err(spk_storage::Error::SPFS(spfs::Error::UnknownRemoteName(_))) => {}
+                Err(err) => return Err(err.into()),
+                Ok(mut origin) => {
+                    if self.legacy_spk_version_tags {
+                        origin.set_legacy_spk_version_tags(true);
```
jrray, line 993: The pin at time section is missing here. I suggest refactoring the loop body above to reuse the same logic.

jrray (changes requested):

crates/spk-cli/common/src/flags.rs
rydrman: Our default origin never had an associated pin time...

jrray: The old code would respect the pin option if given, no?

rydrman: Yes, and for user provided values it still does, but the injected origin repo was always `("origin", None)` so handling the pin in this portion of the code would be redundant

jrray: Okay, I see what you mean. IMO it would still be an improvement to unify the code but I'll approve pending CI passing.

jrray (approved):

crates/spk-cli/common/src/flags.rs
rydrman: Ya, I agree I'll see if there's a trivial way to work a closure in or something

rydrman: I gave this a try but the use of async makes it difficult to use a closure or similar reuse pattern. Happy to have someone iterate though...

jrray: The use of `Option::or` on line 976 means the "origin" repo would still respect the `--when` command line flag, and it is reasonable to assume that `--when` would apply to the default origin repo.

I have a PR on the way to restore this behavior.

PR #1088, opened 2024-08-04
Fix default origin repo not respecting --when

jrray:
As discussed in #1087 reuse the same logic for configuring the implicit "origin" repo as any explicitly enabled repos.

Also fix a problem from the pre-#1087 code where any explicitly enabled "origin" (with custom pin) would be replaced by the implicit one.


rydrman (approved):
Thanks, and kudos for resolving a bug before it had a chance to impact anyone!

PR #1090, opened 2024-08-05
update validation rules for gmp, gcc63 and gcc63

doubleailes:
This PR is meant to update the deprecated `validation` attribut `disable` for the following pckages

- packages/gnu/gmp.spk.yaml
- packages/gnu/gcc/gcc63.spk.yaml
- packages/gnu/gcc/gcc93.spk.yaml

and add `allow` `RecursiveBuild` validation for packages/gnu/gcc/gcc63.spk.yaml

### DIFF
--- a/packages/gnu/gcc/gcc93.spk.yaml
+++ b/packages/gnu/gcc/gcc93.spk.yaml
@@ -40,9 +40,7 @@ build:
     - make install
     # no need to keep extra bloat files in info dir
     - spfs reset /spfs/share/info/**/*
-  validation:
-    rules:
-      - deny: AlterExistingFiles
+
 
 tests:
   - stage: install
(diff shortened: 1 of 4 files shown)


packages/gnu/gcc/gcc93.spk.yaml
```diff
@@ -41,9 +41,8 @@ build:
     # no need to keep extra bloat files in info dir
     - spfs reset /spfs/share/info/**/*
   validation:
-    disabled:
-      # we are replacing the old gcc version on purpose
-      - MustNotAlterExistingFiles
+    rules:
+      - deny: AlterExistingFiles
```
rydrman, line 45: were you able to build these packages successfully? The old rule was disabling the validation, so in the new setup you would need `- allow: AlterExistingFiles` in order to achieve the same behavior

doubleailes:
I remove the validation and the package build but i have other errors.
But i don't have the deprecated anymore.

rydrman:
Thinking about this some more, and we should definitely keep the validation rules, specifically it should become `- allow: AlterExistingFiles` as mentioned before. If you want to submit a new issue with more information about the errors that you are getting locally it might be better to debug that separately to this change

Thanks!

rydrman (approved):

PR #1092, opened 2024-08-06
Fix GCC version to 6.3.0

### LINKED ISSUE #1091: [BUG] GCC 6.3.3 does not exist
**Describe the bug**
The file https://github.com/spkenv/spk/blob/106be3a8be5581e5c57e6b9fea3478fd2ac97994/packages/gnu/gcc/gcc63.spk.yaml#L1 defaults to version 6.3.3 which does not exist at http://ftpmirror.gnu.org/gnu/gcc/ nor in the [GCC releases](https://gcc.gnu.org/releases.html).

**To Reproduce**
1. Go to the root of 'spk'
2. Run `make packages`

**Expected behavior**
Build the `gcc` package without errors.

**Additional context**
While testing to build `spk` with @doubleailes we ran into this issue, but I don't have a build environment myself yet. It seemed legitimate to open a bug anyway, since the fix is simple (based on the commit history).

maximd:
The commit history shows that the GCC version was originally 6.3.0(https://github.com/spkenv/spk/commit/7fb02e3d40596135bee3e58f82799d4f5c677851#diff-633c824ed40d9bb6e821cfa853b3789251fcf4c79857ace47213ebdac3dbbe8c) and the [GCC releases](https://gcc.gnu.org/releases.html) shows 6.3.0 as the only valid 6.3 version.

Fixes #1091


jrray (approved):
Thanks for the fix!

jrray:
It's worth mentioning that at SPI we're using a RHEL-based OS and make use of devtoolset-6 (or other versions) for gcc, and our package recipe looks like:

```yaml
pkg: gcc/6.3.1+r.1

build:
  options:
    - pkg: stdfs
  script:
    - true

install:
  environment:
    - set: CC
      value: /opt/rh/devtoolset-6/root/usr/bin/gcc
    - set: CXX
      value: /opt/rh/devtoolset-6/root/usr/bin/g++
    - prepend: PATH
      value: /opt/rh/devtoolset-6/root/usr/bin
  requirements:
    - pkg: stdfs
```

So we don't actually use a self-compiled gcc and using the devtoolset gcc produces binaries compatible with the base distro. Going this route has the benefit of being able to depend on other libraries provided by the distro without having to make spk packages for things where the distro-provided version all you need.

jrray:
And on Rocky we expose the default gcc this way:

```yaml
# This package is essentially empty -- it represents using the native version
# of gcc installed on rocky 9.3 (gcc 11).
#
# It doesn't attempt to clean the environment ($PATH, $CC, $CXX) which may be
# set to some other compiler.

pkg: gcc/11.4.1

build:
  options:
    - pkg: stdfs
  script:
    - true
  variants:
    - { distro: rocky }

install:
  requirements:
    - pkg: stdfs
```

maximd:
> So we don't actually use a self-compiled gcc and using the devtoolset gcc produces binaries compatible with the base distro. Going this route has the benefit of being able to depend on other libraries provided by the distro without having to make spk packages for things where the distro-provided version all you need.

Ah thanks for the feedback, I figured there must be a way to use the precompiled libraries provided by the OS :)

We are especially interested in testing spk on Rocky 9 as a first step, so thanks for the hint!

Is there any additional documentation on package recipes?

jrray:
We're still figuring out the project configuration now that the repo is public, it seems that this PR didn't run any of the github actions? Anyway CI is failing on the package lint step now that this was merged. I'm investigating.

```
Failed gnu/gcc/gcc63.spk.yaml:
-----> 
   | # 
 2 | pkg: gcc/6.3.O
   |      ^ pkg: Error: expected an ascii digit at O at line 2 column 6
   | api: v0/package
   | 
   | sources:
   |   - tar: http://ftpmirror.gnu.org/gnu/gcc/gcc-6.3.O/gcc-6.3.O.tar.gz
   | 
```

packages/gnu/gcc/gcc63.spk.yaml
```diff
@@ -1,4 +1,4 @@
-# {% set opt = opt | default_opts(version="6.3.3") %}
+# {% set opt = opt | default_opts(version="6.3.O") %}
```
jrray, line 1: This is the letter `O` and not the digit `0` 🙃

maximd: Oh, such a typo is embarrassing, my apologies for the lost time in fixing it.
This is a good reminder that even for quick fixes, local testing is not an option.

Thanks for the fix in https://github.com/spkenv/spk/commit/d9b67b7696fd14a6100c2c072905231c08b6c618 and https://github.com/spkenv/spk/commit/33ab16a1b53d5972f59b420e80e48b0747fb0406 @rydrman @jrray!

I have a question for the future: would it make sense to run the `Linux Build and Test` github action in PR's checks (like the one that failed after the merge: https://github.com/spkenv/spk/actions/runs/10288203486/job/28473074445) or would it [consume too many GitHub actions minutes](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) to be viable?

jrray: > I have a question for the future: would it make sense to run the `Linux Build and Test` github action in PR's checks (like the one that failed after the merge: https://github.com/spkenv/spk/actions/runs/10288203486/job/28473074445) or would it [consume too many GitHub actions minutes](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) to be viable?

Yes, these checks are supposed to run for every PR but we have to manually approve running them for new contributors. As of today we've configured the project so CI needs to pass before a PR can merge.

PR #1101, opened 2024-08-09
Add host filtering options to spk search

dcookspi:
This adds filtering with `--host` and `--no-host` options to `spk search`, like the `ls` command has.


rydrman (approved):

rydrman:
@dcookspi any reason not to merge this?

Issue #1036, opened 2024-06-07
Tracking manifest handling of files named with a dot prefix


jrray:
https://github.com/imageworks/spk/blob/ee915b9a5abd24d39171773f73dd8aced1233b32/crates/spfs/src/tracking/manifest.rs#L230-L232
https://github.com/imageworks/spk/blob/ee915b9a5abd24d39171773f73dd8aced1233b32/crates/spfs/src/tracking/manifest.rs#L270-L273

Are we expecting a path like `/.foo` to be changed into `foo`?

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=fn+main%28%29+%7B%0A++++println%21%28%22%7B%7D%22%2C+%22%2F.foo%22.trim_start_matches%28%26%5B%27%2F%27%2C+%27.%27%5D%29%29%3B%0A%7D)

Was this intended to change `./foo` into `foo`?

PR #1103, opened 2024-08-14
Remove trim_start_matches targeting '.'

jrray:
Use a custom method for trimming '/' or "./" that doesn't change ".foo"
into "foo".

Fixes https://github.com/spkenv/spk/issues/1036

### DIFF
--- a/crates/spfs/src/tracking/diff.rs
+++ b/crates/spfs/src/tracking/diff.rs
@@ -117,22 +117,33 @@ pub fn compute_diff<U1: Clone>(a: &Manifest<U1>, b: &Manifest<U1>) -> Vec<Diff<U
 
     let mut visited = HashSet::new();
     for entry in all_entries.iter() {
-        if visited.contains(&entry.path) {
+        if !visited.insert(&entry.path) {
             continue;
-        } else {
-            visited.insert(&entry.path);
-            match diff_path(a, b, &entry.path) {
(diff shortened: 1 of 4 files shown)


jrray:
I thought this was a problem internal to `Manifest` and made the new private method there, but then found that `spfs-vfs` has the same bug.

crates/spfs/src/tracking/diff.rs
```diff
@@ -117,60 +117,70 @@ pub fn compute_diff<U1: Clone>(a: &Manifest<U1>, b: &Manifest<U1>) -> Vec<Diff<U
...
     changes
 }
 
+// Allow: most instances will be of the large variant; boxing is
+// counter-productive.
+#[allow(clippy::large_enum_variant)]
+enum DiffPathResult<U1, U2> {
+    Diff(Diff<U1, U2>),
+    PathMissingFromBothManifests,
+    UselessMask,
```
rydrman, line 143: I'm not entirely clear on this variant based on the naming, would you mind adding some documentation about when this result appears?

jrray: Did you see "path was masked in the right manifest but didn't exist in the left" above?

That's what I'd add as the description.

rydrman: Oh, yes I see that does help as context

rydrman (approved):

Issue #1105, opened 2024-08-17
Add Nushell support


doubleailes:
**Is your feature request related to a problem? Please describe.**
Using a unify ( win/linux) modern shell with Spk

**Describe the solution you'd like**
An implementation of Nushell in Spk

**Describe alternatives you've considered**
Use powershell also with linux but it not that interesting

**Additional context**
...

doubleailes:
Following the Slack discussion the todo to implement Nushell in Spk should be:

- [x] spfs startup script to initialize the shell inside a new runtime
- [x] spfs activation scripts generated by spk

It generate two issues with Nushell
- [x] [Restriction autoload to only nu files](https://github.com/nushell/nushell/issues/13587)
    - [ ] Released in version 0.97.0
- [ ]  [Autoload directories using --commands](https://github.com/nushell/nushell/issues/13587)
   - [ ]  Released in version ?

Here is a draft PR: https://github.com/spkenv/spk/pull/1106 to start discussion waiting Nushell team to add the fix/features

doubleailes:
The PR https://github.com/nushell/nushell/pull/15679 add to Nushell the required feature.

PR #1106, opened 2024-08-17
Add nushell

doubleailes:
Here is a draft PR to start discussion waiting Nushell team to add the fix/features link in the issue: https://github.com/spkenv/spk/issues/1105


rydrman:
Looks like great progress, let us know if you need any other support while you wait for the necessary nu features

PR #1107, opened 2024-08-19
Add repo name to 'opening repository' debug messages

dcookspi:
This adds the repo's name to the 'opening repository ... ' debug messages to make each mesasge clearer. It can be tricky to work out which repo each message refers to when you have complicated multi-repo configurations.

### DIFF
--- a/crates/spfs/src/config.rs
+++ b/crates/spfs/src/config.rs
@@ -587,11 +587,19 @@ impl Config {
         match self.remote.get(remote_name.as_ref()) {
             Some(Remote::Address(remote)) => {
                 let config = RemoteConfig::from_address(remote.address.clone()).await?;
-                tracing::debug!(?config, "opening '{}' repository", remote_name.as_ref());
+                tracing::debug!(
+                    ?config,
+                    "opening '{}' repository via address",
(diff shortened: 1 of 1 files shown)


crates/spfs/src/config.rs
```diff
@@ -587,11 +587,11 @@ impl Config {
         match self.remote.get(remote_name.as_ref()) {
             Some(Remote::Address(remote)) => {
                 let config = RemoteConfig::from_address(remote.address.clone()).await?;
-                tracing::debug!(?config, "opening repository");
+                tracing::debug!(?config, "opening '{}' repository", remote_name.as_ref());
```
jrray, line 590:
```suggestion
                tracing::debug!(?config, "opening '{}' repository via address", remote_name.as_ref());
```

This change looks fine to me, but there's also an opportunity here to disambiguate these two trace lines so we can tell which path was taken from the log. The other one could say "via config" for symmetry.

jrray (approved):

crates/spfs/src/config.rs
dcookspi: I've added different text to each log message.

PR #1108, opened 2024-08-20
default runtime dir use temp_dir

doubleailes:
This PR remove `cont RUNTIME_DIR` to use `temp_dir()` and support windows temp dir


crates/spfs/src/runtime/storage.rs
```diff
@@ -408,12 +408,12 @@ pub struct Config {
 
 impl Default for Config {
     fn default() -> Self {
-        Self::from_root(Path::new(Self::RUNTIME_DIR))
+        // Default RUNTIME_DIR
+        Self::from_root(temp_dir().join("spfs-runtime"))
```
rydrman, line 412:
```suggestion
        let default_runtime_dir = temp_dir().join("spfs-runtime");
        Self::from_root(default_runtime_dir)
```

A named variable allows us to avoid the declarative comment :)

doubleailes:
Sorry i did not comment the suggestion but i accepted it.

cheers.

rydrman (approved):

rydrman:
@jrray FYI as I'm not sure if you use a separate temp dir at SPI. I will double check at ILM as we do but I don't think this has any negative impact

jrray:
> @jrray FYI as I'm not sure if you use a separate temp dir at SPI. I will double check at ILM as we do but I don't think this has any negative impact 

On Linux we're mounting a tmpfs here, in the private mount namespace, right? I want to say that means the path doesn't matter too much.

doubleailes:
I rebase to get the https://github.com/spkenv/spk/pull/1119 changes

doubleailes:
Since test are running i need to look deeper into it. Move back to draft.

PR #1110, opened 2024-08-20
add win interruption using tokio::windows

doubleailes:
This PR add support for Windows interruption signal using ctrl-c based on `tokio::signal::windows::ctrl_c`

I alternate unix/win to streamline comparaison but i can reorg to regroup by OS.


crates/spfs-cli/cmd-monitor/src/cmd_monitor.rs
```diff
@@ -86,11 +89,10 @@ impl CmdMonitor {
         rt.block_on(self.wait_for_ready());
         // clean up this runtime and all other threads before detaching
         drop(rt);
-
+        #[cfg(unix)]
```
rydrman, line 92: Is this working for you? I think the monitor still needs to be background-ed to not block the runtime from launching, no?

crates/spfs-cli/cmd-monitor/src/cmd_monitor.rs
```diff
@@ -142,12 +144,23 @@ impl CmdMonitor {
...
         let mut quit = signal(SignalKind::quit())
             .map_err(|err| Error::process_spawn_error("signal()", err, None))?;
+        #[cfg(windows)]
+        let mut quit = ctrl_c().map_err(|err| Error::process_spawn_error("ctrl_c()", err, None))?;
+        #[cfg(unix)]
         let mut terminate = signal(SignalKind::terminate())
             .map_err(|err| Error::process_spawn_error("signal()", err, None))?;
+        #[cfg(windows)]
+        let mut terminate =
+            ctrl_c().map_err(|err| Error::process_spawn_error("ctrl_c()", err, None))?;
```
rydrman, line 163: To avoid all this `#[cfg]` maybe we can create a single future to wrap all the signals - 

specifically, define a trait with a function that builds a future for all signals in one that can be implemented differently on this `CmdMonitor` type for each os. Let me know if my description is not clear

crates/spfs-cli/cmd-monitor/src/cmd_monitor.rs
doubleailes: I guess you mean under windows?

So far i was able to trig the commands of course with error but not depending of the runtime blocking

doubleailes:
I switch back to draft. I need further test.

Issue #895, opened 2023-10-19
spfs-fuse process can get left behind


rydrman:
This is a bit of an edge case but wanted it documented nonetheless.

We had a host which was failing to start an spfs shell upon login via ssh. In this case the user would see this error (overlayfs + fuse):

```
WARN FUSE did not appear to start after delay: Connection refused (os error 111)                                                                                                      
mount: mount none on /spfs failed: Connection refused                                                                                                                                  
ERROR Failed to mount overlayfs
```

A look at the system journal would show this for spfs-fuse:

```
spfs[4439]:  INFO Filesystem initialized
spfs[4439]:  WARN Request RequestId(1): Failed to send reply: Invalid argument (os error 22)
```

In this case, spfs-fuse was running but spfs-enter failed because overlayfs couldn't be mounted. This meant that the monitor was never started and the spfs-fuse process would stick around forever.

I was not able to identify the underlying fuse issue, and rebooting the machine resolved the mount error so we moved on. 

This issue is to try and track the failure state, and have a way in which these partial mounts can still be properly torn down on failure.

jrray:
https://github.com/spkenv/spk/blob/8151a88667f9fd4968ddfdaa5a93616a28c22c07/crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs#L272-L275

FWIW I discovered in the documentation for `abort` that it doesn't work on join handles returned from `spawn_blocking`. Despite the comment attached here, with `abort` being a no-op we really weren't doing any kind of cleanup here. In my testing, using fuser's `unmount` method does nothing and there's no way to signal to `fuser::Session::run`'s loop to terminate.

I've implemented a heartbeat connection between spfs-monitor and spfs-fuse, as suggested above (and recently in slack). Before adding this heartbeat, it is easily reproducible to get a spfs-fuse process hanging around forever by `kill -9`'ing the related spfs-monitor process. But with the heartbeat in place, spfs-fuse will eventually timeout and exit.

PR #1111, opened 2024-08-21
Add a heartbeat between spfs-monitor and spfs-fuse

jrray:
As mentioned in #895, there are cases where spfs-fuse never shuts down. An easy way to repro this is to `kill -9` the spfs-monitor process, so it doesn't get a chance to clean up the runtime normally.

We observe spfs-fuse processes accumulating over time on our CI runners. It could be from users canceling pipelines (maybe the runner does a `kill -9` of all the child processes?). It is also extremely common for zombie runtimes to accumulate over time (not just on CI runners), perhaps spfs-monitor is crashing / failing to cleanup commonly?

~~This is a proof of concept and enabling this or tuning the timing of the heartbeat and timeout should be a configurable thing but these are hard coded for now.~~ Configuration added!

### DIFF
--- a/crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
+++ b/crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
@@ -10,7 +10,7 @@ use fuser::MountOption;
 use miette::{bail, miette, Context, IntoDiagnostic, Result};
 use spfs::tracking::EnvSpec;
 use spfs::Error;
-use spfs_cli_common as cli;
+use spfs_cli_common::{self as cli, warn_and_sentry_event};
 use spfs_vfs::{Config, Session};
 use tokio::signal::unix::{signal, SignalKind};
 use tokio::time::timeout;
@@ -275,7 +275,8 @@ impl CmdFuse {
                         let seconds = session.seconds_since_last_heartbeat();
                         tracing::trace!(seconds_since_last_heartbeat = ?seconds, "heartbeat monitor");
                         if seconds > heartbeat_grace_period_seconds {
-                            tracing::warn!("loss of heartbeat, shutting down filesystem");
+                            warn_and_sentry_event!("loss of heartbeat, shutting down filesystem");
+
                             // XXX: Calling unmount here has no apparent effect!
                             heartbeat_send.send(()).await.into_diagnostic().wrap_err("Failed to send unmount signal")?;
                             break;
--- a/crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
+++ b/crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
@@ -3,6 +3,7 @@
 // https://github.com/spkenv/spk
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use clap::Parser;
 use fuser::MountOption;
@@ -12,6 +13,7 @@ use spfs::Error;
 use spfs_cli_common as cli;
 use spfs_vfs::{Config, Session};
 use tokio::signal::unix::{signal, SignalKind};
+use tokio::time::timeout;
 
 // The runtime setup process manages the current namespace
 // which operates only on the current thread. For this reason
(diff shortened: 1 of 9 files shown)


crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
```diff
@@ -268,11 +289,40 @@ impl CmdFuse {
...
+            // cleanup.
+            if heartbeat_failed {
+                return res;
+            }
+            if !join_handle.is_finished() {
+                // XXX: Calling unmount has no apparent effect!
+                unmount_callable.lock().unwrap().unmount().into_diagnostic().wrap_err("FUSE unmount failed")?;
+                tracing::trace!("Joining FUSE session");
+                join_handle.await.into_diagnostic().wrap_err("FUSE join_handle await failed")?.into_diagnostic().wrap_err("FUSE session failed after unmount")?;
+                tracing::trace!("FUSE session joined");
```
jrray, line 324: This is meant to shutdown cleanly, but in practice it will park forever waiting on the join_handle. In order for the run loop to exit, the fuse filesystem needs to be unmounted, but the unmount call here doesn't work. As mentioned in the comment, even fusermount can apparently succeed at unmounting the path, but the spfs filesystem continues to function normally when it is part of an overlayfs stack.

On the other hand, just returning and allowing the process to terminate will allow the fusermount auto_unmount to kick in. At least in my testing, after `kill -9`'ing spfs-monitor and allowing the heartbeat to expire, spfs-fuse exits and the accompanying fusermount process goes away.

Attempting to access /spfs after that point does this:
```
jrray@STRANDED:~/git/spk$ ls /spfs
ls: reading directory '/spfs': Transport endpoint is not connected
```

(This is in my ubuntu home environment.)

jrray:
I think the biggest concern is linking the lifetime of spfs-fuse to spfs-monitor. At face value that's what we want in order to solve the zombie spfs-fuse problem, but if spfs-monitor is crashing unexpectedly which the runtime is still in use then the fuse filesystem would be shutting down prematurely.

If spfs-monitor is crashing that's a serious problem that needs to be sorted out. This heartbeat link may be dangerous until that happens. But is it crashing, or is it getting killed? This is unclear. If it is getting killed (with -9) because a job is killed in CI or on the farm or whatever, then before this change spfs-fuse would keep running forever, and after this change it would terminate.

crates/spfs-vfs/src/fuse.rs
```diff
@@ -790,6 +807,14 @@ impl fuser::Filesystem for Session {
     }
 
     fn lookup(&mut self, _req: &Request<'_>, parent: u64, name: &OsStr, reply: ReplyEntry) {
+        if name.as_bytes().starts_with(b".fuse-heartbeat") {
+            let seconds_since_session_start = self.inner.session_start.elapsed().as_secs();
+            tracing::trace!(?seconds_since_session_start, "heard heartbeat");
+            self.inner
+                .last_heartbeat_seconds_since_session_start
+                .store(seconds_since_session_start, Ordering::Relaxed);
+        }
```
jrray, line 816: Note that this behavior is a side-effect and the lookup is still processed normally. This could be considered "expensive" overhead that has to communicate with remote spfs repos to process.

The magic filename pattern could include a hard-coded uuid to make it "impossible" that /spfs environment would actually contain this file, and then this code could synthesize the reply itself without processing the lookup normally.

crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
jrray: I added a timeout to the join.

crates/spfs-vfs/src/fuse.rs
jrray: I created a "magic" filename prefix.

crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
```diff
@@ -268,11 +294,52 @@ impl CmdFuse {
...
+                return res;
+            }
+            if !join_handle.is_finished() {
+                // XXX: Calling unmount has no apparent effect!
+                unmount_callable.lock().unwrap().unmount().into_diagnostic().wrap_err("FUSE unmount failed")?;
+                tracing::trace!("Joining FUSE session");
+                // Since the umount above may have no effect, this join uses a
+                // timeout so the process doesn't deadlock. Once this process
+                // exits, the fusermount auto_unmount should kick in.
+                match timeout(Duration::from_secs(5), join_handle).await {
```
dcookspi, line 333: Is it worth being able to configure this timeout?

dcookspi (approved):

crates/spfs-cli/cmd-fuse/src/cmd_fuse.rs
jrray: From my testing no amount of waiting will ever succeed, so making this configurable might just mislead people into thinking it could matter. Except maybe to let it be configured shorter than 5 seconds?

I wanted to leave this code that attempts to clean up gracefully in place although it isn't working properly, and the timeout is a concession to leave the code in place without hanging forever. Hopefully this can be revisited and figured out later.

rydrman (approved):

Issue #920, opened 2023-11-24
SPFS Configurable Tag Lock Expiry


rydrman:
Stale tag locks can get left behind in many scenarios, and cause problems in a repo because they block operations and are difficult to clean up safely. I propose adding a configurable lifetime to all tags, with the assumption that writing a tag file should only take seconds at most - although there are edge cases to this assumption, I think it could be worth the trade-ff to have spfs check the age of a lock and forcefully take it over if it's beyond the configured threshold. We could set a seemingly reasonable default like 60s.

A possible solution to #623

PR #1117, opened 2024-08-27
Add method to unlock tags manually

rydrman:
Very beginning of #920 that we are using internally to programmatically clear tag locks in some cases.

### DIFF
--- a/crates/spfs/src/storage/fs/tag.rs
+++ b/crates/spfs/src/storage/fs/tag.rs
@@ -130,6 +130,29 @@ impl TagStorage for FsRepository {
     }
 }
 
+impl FsRepository {
+    /// Forcefully remove any lock file for the identified tag.
+    ///
+    /// # Safety
+    /// This function is unsafe because it removes the lock file without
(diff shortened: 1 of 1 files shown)


crates/spfs/src/storage/fs/tag.rs
```diff
@@ -883,6 +923,23 @@ impl TagLock {
...
+
+    /// Remove the lock file for a tag
+    ///
+    /// # Safety:
+    /// Tag locks are used to ensure that only one process is writing to a tag file at a time.
+    /// Removing the lock file without ensuring that the tag file is not being written to may
+    /// cause the data within the file to become corrupt.
+    pub unsafe fn remove<P: AsRef<Path>>(tag_file: P) -> Result<()> {
+        let mut lock_file = tag_file.as_ref().to_path_buf();
+        lock_file.set_extension("tag.lock");
```
jrray, line 935: Now that this `"tag.lock"` appears in multiple places I suggest moving the value to a const.

jrray (approved):

PR #1124, opened 2024-09-11
Adds support for parsing a comma-separated list of range idents

dcookspi:
This adds an api function for parsing a comma-separated list (csl) of range idents (packages or requests). It required an update to the parser for `package_name_not_version`, and it includes some intial proptests for parsing valid and invalid range ident comma-separated lists.

The background is that SPI had a csl parser for package requests in another tool that uses `spk` as a library. `spk` parsed the package identifiers, and the tool dealt with the list. We hit issue parsing these csl's when some of the requests contained post-releases or multiple version ranges, e.g. `pkgA/1,mypkg/2.3.1+r4,pkgB` or `mypkg/1.2.3,1.3.5,pkgC`. 

This led to the `package_name_not_version` parser. We found it wasn't parsing the packages correctly in these situations and could mistake package names for repo names, e.g. `mypkg` in `mypkg/1.2.3,pkgD` would be treated as a package called `1.2.3` in the repo `mypkg`.  As part of fixing that, we wanted to move the csl parsing into `spk`, and add tests for it, So it could be used in future work and all the `spk` package/request/range ident parsing is kept together.

### DIFF
--- a/crates/spk-schema/crates/foundation/src/ident_ops/parsing/mod.rs
+++ b/crates/spk-schema/crates/foundation/src/ident_ops/parsing/mod.rs
@@ -67,7 +67,7 @@ where
         // package name instead of as a repository name.
         let r = version_parser
             .parse(input)
-            .and_then(|(input, _)| alt((tag("]"), eof::<&str, _>))(input));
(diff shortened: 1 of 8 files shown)


jrray (approved):
LGTM!

PR #1125, opened 2024-09-16
Make sync progress bars optional

jrray:
The goal is to have the ability for commands like `spfs run` and `spk env` to not show a progress bar when running the `Syncer`. A new `--no-progress-bars` flag has been added where appropriate.

The `Syncer` type is no longer generic over the `SyncReporter` in order for `Sync::get_syncer` to be able to return a `Syncer` with either reporter type. The existing code was not doing anything useful with it being generic over the `SyncReporter`.

Some `*_with_reporter` alternate functions have been added to spk-exec since these functions were hard-coded to show progress bars.


crates/spfs/src/sync.rs
```diff
@@ -69,13 +69,98 @@ impl SyncPolicy {
...
+        (**self).visit_payload(digest)
+    }
+    fn synced_payload(&self, result: &SyncPayloadResult) {
+        (**self).synced_payload(result)
+    }
+}
+
+#[derive(Clone)]
+#[enum_dispatch::enum_dispatch(SyncReporter)]
+pub enum SyncReporters {
```
rydrman, line 140: Our downstream codebase relies on a custom sync reporter implementation - maybe we `Box` up the trait instead of moving to an enum?

jrray: I added a `Custom` variant to the enum.

rydrman (approved):

PR #1126, opened 2024-09-16
Create concrete types for Compatibility::Incompatible

jrray:
This was a large undertaking to make progress towards improving solver error messages. Remove `IncompatibleReason::Other` which was being used as a catchall and add enum variants for all the different kinds of reasons. This was done primarily so that these reasons can be compared to possibly merged. For example, when all the available versions of a package have a version number too low to satisfy a requests, all the `TRY` lines can be replaced with a single note that says that all the versions are too low.

Care was taken to preserve the way the reasons are presented as a string (with some minor tweaks). To accomplish this, the enum variants need to carry additional data but also know what parts of that data factor into if two instances of the same variant are considered the same reason. A new trait with the clunky name `IsSameReasonAs` serves as the mechanism for comparing reasons. A convention of `*Problem` type names is used to support the cases where a single `IncompatibleReason` variant can have multiple different strings depending on context.

It is hoped that with these changes the memory footprint of the solver is lowered since the solver no longer has to proactively create strings when it was producing `IncompatibleReason::Other` instances; the strings are created on demand if they are needed when the solver output is being generated. However the new `IncompatibleReason` variants need to clone various things so they still have a memory footprint. An attempt was made to add a generic lifetime to `Incompatible` to be able to just keep references around but it resulted in some seemly intractable borrow checker problems.


rydrman (approved):
I'll be honest, I scanned a lot of this but certainly didn't comb through any of it. I think that this is a net-positive and the spirit seems to not disrupt any existing logic that would cause bugs. I'm happy to get this merged as-is and avoid letting conflicts build up against the changes.

PR #1129, opened 2024-09-18
Preserve install environment

### LINKED ISSUE #1081: `install.environment` config is lost when recipe is published
Given a recipe with `install.environment` rules:

```yaml
pkg: env/1.0.0

api: v0/package

build:
  script:
    - true

install:
  environment:
    - set: TEST
      value: test
```

After building this package, the recipe that is created has lost these rules:

```sh
$ spk info env/1.0.0
api: v0/package
pkg: env/1.0.0
sources:
- path: .
  exclude:
  - .git/
  - .svn/
  filter:
  - :- .gitignore
build:
  script:
  - 'true'
```

(`$ spfs read spk/spec/env/1.0.0` gives the same output.)

The rules can still be found in the `src` recipe (`$ spk info env/1.0.0/src`) but it would be better if they survived into the other recipes for visibility.

jrray:
Refactor various is_default() methods into IsDefault trait

Fixes #1081. The bug causing #1081 was in `InstallSpec::is_default` as the
logic was not updated when the `environment` field was added to the struct
way back at https://github.com/spkenv/spk/blob/f4899ce1e4ac0634ee95608e322a86a358096d4d/src/api/install_spec.rs#L31

As a guard against this kind of bug in the future, an `IsDefault` trait was
created and a derive macro created to derive the `is_default` method
semi-automatically. All existing `is_default` methods have been changed to
use the trait instead.

And...

[Refactor Package::runtime_environment](https://github.com/spkenv/spk/pull/1129/commits/523943e60f764d76bff1e5e0697f94e90f64f4e6)

Both recipes and packages have the `install.environment` property, so
refactor `runtime_environment()` out of the `Package` trait into its own
trait, then require both `Recipe` and `Package` implement it.

This change was prompted in order to make the code in the new test possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1825,6 +1825,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7655c9839580ee829dfacba1d1278c2b7883e50a277ff7541299489d6bdfdc45"
 
 [[package]]
+name = "is_default_derive_macro"
+version = "0.42.0"
+dependencies = [
+ "indicatif",
+ "quote",
+ "syn 2.0.48",
+ "tracing",
+]
+
+[[package]]
 name = "itertools"
 version = "0.10.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


rydrman (approved):
Nice, love the addition of the proc macro as well

PR #1133, opened 2024-09-27
Fix empty Request requester in test requirements

jrray:
Any package requirements on an spk recipe test need to have their
requester filled in.

    -RESOLVE python-pytest/7.4.3/UNZ5Q23T  (requested by )
    +RESOLVE python-pytest/7.4.3/UNZ5Q23T  (requested by some-package-name/1.2.1 install test)

### DIFF
--- a/crates/spk-schema/src/v0/spec.rs
+++ b/crates/spk-schema/src/v0/spec.rs
@@ -23,7 +23,7 @@ use spk_schema_foundation::version::{
     VarOptionProblem,
 };
 use spk_schema_foundation::IsDefault;
-use spk_schema_ident::{AnyIdent, BuildIdent, Ident, RangeIdent, VersionIdent};
+use spk_schema_ident::{AnyIdent, AsVersionIdent, BuildIdent, Ident, RangeIdent, VersionIdent};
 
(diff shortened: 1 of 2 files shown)


crates/spk-schema/crates/ident/src/ident.rs
```diff
@@ -201,3 +203,16 @@ where
...
+
+/// Idents that can be turned into a [`VersionIdent`] can implement this trait.
+pub trait AsVersion {
+    /// Return a borrowed version of this ident converted into a
+    /// [`VersionIdent`].
+    fn as_version(&self) -> &VersionIdent;
+}
+
+impl<T> AsVersion for Ident<VersionIdent, T> {
+    fn as_version(&self) -> &VersionIdent {
```
rydrman, line 215: Any thoughts about this being `AsVersionIdent::as_version_ident`? I only worry that it sounds like it should be just the version as a result

rydrman (approved):

crates/spk-schema/crates/ident/src/ident.rs
jrray: There was already `as_version` methods on the idents before this change, so I took the nuclear option in #1134.

PR #1136, opened 2024-10-03
Add list of layers file to spfs run command

dcookspi:
This adds support for list of layer references file to spfs run's command line to allow for larger environments (number of layesr) in smaller/shorter command lines.

This changes the `EnvSpecItem` for live layer files to allow more than one kind of yaml file in environment specifiers. It adds an EnvLayersFile that can contain a list of spfs references: disgests, tags, etc.. This extends and reworks the parsing and enums to support the new file, and moves LiveLayers to their own source code file.

Note: I'm not sure about some of the names I used for enums and entries, and the location of at least one enum. See the `TODO:` markers in the code, suggestions welcome.

An example layers list file:
```yaml
# layers.spfs.yaml
api: v0/layerlist
layers:
  - A7USTIBXPXHMD5CYEIIOBMFLM3X77ESVR3WAUXQ7XQQGTHKH7DMQ====
  - spfs/some/tag/to/some/object
  - /path/to/some/live_layer.spfs.yaml
  - 6PJDUUENJYFFDKZWRKHDUXK4FGGZ7FHDYMZ7P4CXORUG6TUMDJ7A====
```
**Note:**
I haven't put any checks in to stop references to other layer list files in the code, or any depth limits on them. So it should be possible to put paths to other layer list files in a layer list file, and nest them indefinitely. I haven't tried that. If that's a concern I can add a check or limit.


dcookspi:
Todo: 
- [ ] remove teh `spfs` from the api verion mapping related types
- [ ] decide on references to layer files inside layer files, currently a recursive reference will break spfs run

crates/spfs/src/tracking/env.rs
```diff
@@ -19,13 +26,199 @@ pub const ENV_SPEC_SEPARATOR: &str = "+";
...
 pub const ENV_SPEC_EMPTY: &str = "-";
 
+// For the config file that contains the extra bind mounts
+const SPFS_FILE_SUFFIX_YAML: &str = ".spfs.yaml";
+const DEFAULT_LIVE_LAYER_FILENAME: &str = "layer.spfs.yaml";
+
+// For preventing recursive loops when reading list of layers spfs
+// spec files from the command line
+static SEEN_SPEC_FILES: Lazy<std::sync::Mutex<HashSet<PathBuf>>> =
+    Lazy::new(|| std::sync::Mutex::new(HashSet::new()));
```
rydrman, line 36: It doesn't look like this ever gets cleared, right? We have some long-running processes and I worry that we might even simply try to load the same file twice which would cause this to error with no way to recover. It's something that I can refactor if we run into it, but for the future I think we should avoid this kind of global state unless it actually lives in the command line crate.

crates/spfs/src/runtime/spec_api_version.rs
```diff
@@ -0,0 +1,20 @@
...
+// https://github.com/spkenv/spk
+
+use serde::{Deserialize, Serialize};
+
+/// Supported spfs spec api versions
+#[derive(Debug, Deserialize, Serialize, Copy, Clone, Eq, PartialEq, strum::Display)]
+pub enum SpecApiVersion {
+    #[serde(rename = "v0/layer", alias = "v0/livelayer")]
+    V0Layer,
+    #[serde(rename = "v0/layerlist")]
```
rydrman, line 12: my final hesitation with `layerlist` is that it feels too explicit - eg we won't be able to extend the data in this file without the name becoming confusing. Did we already consider `v0/platform`? Because I could also see this being a really useful way to define and then commit platforms directly

crates/spfs/src/tracking/env.rs
dcookspi: Yes, it's not cleared. I've added a function to clear it, in case we need it. I did feel while working on this, the things in `tracking/env.rs` seemed to be partly closer to command line level (parsing), and partly lower level (tag/digest resolution).

crates/spfs/src/runtime/spec_api_version.rs
dcookspi: It is closer to an EnvSpec. It's basically anEnvSpec in a file. How about `v0/envspec`? 
Or `v0/referencelist` to line up with what `spfs run -h` calls an env spec (`REFERENCE`).

jrray: `v0/runspec`!

crates/spfs/src/runtime/spec_api_version.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use serde::{Deserialize, Serialize};
+
+/// Supported spfs spec api versions
+#[derive(Debug, Deserialize, Serialize, Copy, Clone, Eq, PartialEq, strum::Display)]
+pub enum SpecApiVersion {
+    #[serde(
+        rename = "spfs/v0/livelayer",
+        alias = "v0/livelayer",
+        alias = "v0/layer"
```
rydrman, line 13: Feel free to remove these aliases unless you have a lot of these files lying around already - it seems better if we don't need to support these unqualified ones

rydrman (approved):

dcookspi:
I needed to update the docs the for `spfs/v0/runspec` change. So that's done now.

PR #1138, opened 2024-10-11
Bugfix: `get_unresolved_requests()` and incompatible requests

dcookspi:
This fixes a bug in `get_unresolved_requests()` where it would use `unwrap` instead of checking the result. This caused spk to panic when given requests like `mypkg/==1.0.99 mypkg/=1.0.33+r.1`.

The fix removes the `unwrap`, and changes the method to return Result for better error handling. The method needed to be reworked a bit because of the caching mechanism used to avoid recalculating unresolved requests.

### DIFF
--- a/crates/spk-solve/crates/graph/src/graph.rs
+++ b/crates/spk-solve/crates/graph/src/graph.rs
@@ -1375,25 +1375,20 @@ impl State {
     /// Get a mapping of pkg name -> merged request for the unresolved
     /// PkgRequests in this state
     pub fn get_unresolved_requests(&self) -> Result<&HashMap<PkgNameBuf, PkgRequest>> {
-        match self.cached_unresolved_pkg_requests.get() {
-            Some(hm) => Ok(hm),
-            None => {
-                let mut unresolved: HashMap<PkgNameBuf, PkgRequest> = HashMap::new();
+        self.cached_unresolved_pkg_requests.get_or_try_init(|| {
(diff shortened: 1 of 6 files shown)


crates/spk-solve/crates/graph/src/graph.rs
```diff
@@ -1374,23 +1374,26 @@ impl State {
...
-                }
-                if self.get_current_resolve(&req.pkg.name).is_err() {
-                    unresolved.insert(
-                        req.pkg.name.clone(),
-                        self.get_merged_request(&req.pkg.name).unwrap(),
-                    );
+    pub fn get_unresolved_requests(&self) -> Result<&HashMap<PkgNameBuf, PkgRequest>> {
+        match self.cached_unresolved_pkg_requests.get() {
+            Some(hm) => Ok(hm),
+            None => {
```
rydrman, line 1380: could you not use `get_or_try_init` instead of this additional matching?

dcookspi: It's marked as nightly experimental. I tend to avoid those. But I've updated this to use it.

rydrman: in the `std::cell` implementation, yes but we are using the `once_cell` crate in which it's stable - otherwise, I would agree to avoid it

rydrman (approved):

PR #1142, opened 2024-10-29
Add taplo configuration file

jrray:
For formatting Cargo.toml according to our prevailing standards (for those using it in their IDEs).

### DIFF
--- a/.taplo.toml
+++ b/.taplo.toml
@@ -1,0 +1,2 @@
+[formatting]
+indent_string = "    "


rydrman (approved):
I don't know taplo, but feels reasonable

jrray:
I've had this as an uncommitted file in my clone forever but now that I'm trying out git worktrees it will be convenient to have it included in the project.

PR #1148, opened 2024-11-23
Make environment variables to preserve configurable

jrray:
We were preserving $TMPDIR specfically in a hard-coded way and there was no way to opt out of it if desired. Now it has come to light that $LD_LIBRARY_PATH is also cleared when entering an spfs runtime. We would like to preserve that value too. Add a new configuration option so each site can decide to opt in to the preserving. Doing so may come with risks of privilege escalation.


jrray:
Didn't mean to close this. Too many tabs open...

rydrman (approved):

PR #1155, opened 2024-11-27
Make Repository dyn-compatible

jrray:
This involves moving the method that has generics into a separate trait,
and then by extension traits that needed that method into their own
separate traits.

Now it is possible to use `dyn Repository` which may become useful.


jrray:
Here's some literature on the convention of using `FooExt` to name a trait that is used for non-object-safe methods of trait `Foo`: https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

rydrman (approved):

PR #1157, opened 2024-12-02
Refactor renderer for win

jrray:
The main goal here was to remove much of the duplicate code in renderer_unix.rs and renderer_win.rs, but it ended up being a general windows build cleanup.

It is a challenge working in linux and not getting any indication that the windows build is broken until waiting for CI. And then it is an additional challenge trying to fix the windows build while working on a linux system. I was able to get a windows dev environment working and fixed all the build errors and lints that showed up there.

The Rust update to 1.83.0 also brought some new lints with it.


crates/spfs/src/bootstrap.rs
```diff
@@ -275,6 +277,8 @@ pub(crate) fn build_spfs_remove_durable_command(
     })
 }
 
+// Allow: not used on Windows.
+#[allow(dead_code)]
```
rydrman, line 281: should these not just be `#[cfg(unix)]` instead of allowing the dead code?

rydrman (approved):

PR #1158, opened 2024-12-05
Refactor rendering logic into separate methods

jrray:
This change was prompted by wanting to refactor the code to work towards
having different trait bounds for the functions that need a repository
that supports renders (e.g., rendering into a repository) versus
functions that render into a plain directory.

Rework `RenderType` so the method that renders into a hardlink doesn't
need to worry about getting a `RenderType::Copy` value.

### DIFF
--- a/crates/spfs-cli/cmd-render/src/cmd_render.rs
+++ b/crates/spfs-cli/cmd-render/src/cmd_render.rs
@@ -31,10 +31,10 @@ pub struct CmdRender {
     /// using a local directory and `HardLink` for the repository.
     #[clap(
         long,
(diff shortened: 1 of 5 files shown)


jrray:
@rydrman do you have a use case for hard linking into external directories?

rydrman:
Yes, we leverage this quite extensively. Our storage is read-only for users and the permissions are standardized so we hardlink nearly all of our large deployments

jrray:
Okay I guess I need to rethink my refactoring. However don't you agree that this is a dangerous feature to support, in general? I think it might be good to gate it behind some kind of opt-in setting or something, so people understand the consequences.

rydrman:
A warning or extra confirmation on the command line maybe since that's really the only place you can invoke it like that...

rydrman (approved):

PR #1159, opened 2024-12-07
Refactor FsRepository's renders

jrray:
The goal is to make it so it is possible to open repo without creating the renders directory for the current user, in situations where that renders directory will not be used. Examples include when using fuse or when running `spfs clean`.

This introduces different flavors of a "render store" besides `RenderStore` that either represent a render store that hasn't been created yet (but can be) or the lack of a render store. Operations that don't require a render store can access the local repository without the renders directory being created.

WIP: The test suite passes with this code but it is unfinished.

### DIFF
--- a/crates/spfs-cli/cmd-render/src/cmd_render.rs
+++ b/crates/spfs-cli/cmd-render/src/cmd_render.rs
@@ -4,9 +4,10 @@
 
 use clap::builder::TypedValueParser;
 use clap::Parser;
-use miette::{Context, Result};
+use miette::{Context, IntoDiagnostic, Result};
 use spfs::prelude::*;
 use spfs::storage::fallback::FallbackProxy;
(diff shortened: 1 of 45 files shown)


jrray:
Tests are now passing legitimately (at least locally, with an spfs installed from this code). What's left besides general cleanup is creating some new tests that specifically exercise the desired outcome: not creating the renders directory if using fuse. I'm not sure if that is true yet with these changes but the framework for making it possible should be there.

jrray:
Making progress but the behavior of `spfs clean` is not right yet (it currently skipping cleaning renders entirely). The existing tests via the fixture create a repo with `RenderStore` (pre-created renders) but `spfs renders` uses `MaybeRenderStore`, then the clean code treats this flavor as not having renders.

I'm trying to get it so `spfs clean` does clean any existing renders that are eligible without creating any new renders directories.

Issue #1121, opened 2024-09-04
Local Monorepo Build Graph


rydrman:
**Is your feature request related to a problem? Please describe.**

At ILM, we have a distinct workflow where many packages are maintained from a single repo (not currently spk). Our developer and CI workflows operate on "collections" that construct a simple dependency DAG and then build packages that don't exist in order. Notably, the package recipes work for all possible versions of the package and so the version is provided by the collection, not the package spec.

**Describe the solution you'd like**

We would like to convert these packages into spk packages and then have a way to orchestrate builds in the same way using spk.

**Describe alternatives you've considered**

We could hook spk into our existing system, eg build the DAG ourselves and then invoke spk, but resolving the dependencies properly in such a DAG is difficult due to spk spec files being difficult to introspect from the outside, especially as a set of var options are included in the picture.

rydrman:
From the meeting today:
- related to #1120 in that we could possibly do this with more useful introspection of spk recipes
- also related to other issues orchestrating variant builds in CI and properly determining what needs to be done
- we kind of have this if we do build from source with published packages, though there is no great way to interpret the output of `spk explain`
  - this can't handle versions, so we could instead inject the version in order to create the source packages and then run spk
- this is also related to the package specs and build process that we have in the spk repo itself. This is a good first use case for related functionality - having it be able to build the tree on its own without Make.

rydrman:
From the meeting today:
- a "workspace" concept to allow spec discovery and some top-level configuration
- extension of the existing platform spec to better support build-time vs run-time usage
  - at build time you either put a request or specify an exact package version (in case the workspace has templated recipes that you want to build)
  - at run time being able to leverage components to have different requests from the build ones (build is usually lowest-supportable, run is usually highest-compatible)

rydrman:
Here's a WIP proposal for the yaml structure to keep the conversation going

```yaml
api: v0/workspace

# a set of package recipes can be used to build new versions
# of packages as needed by the orchestration process
recipes: 
  # either globs or individual paths can be used to collect recipes
  - packages/*/*.spk.yml
  # these can either be platform recipes or package recipes that
  # might be referenced by builds in the workspace
  - platforms/*/*.spk.yml
```

```yaml
api: v1/platform
platform: dcc-platform
base:
    # the ilm workflow requires inheriting from multiple bases
    # so we update to support a list in this position, order matters
    - company-platform
    - exr-platform

# the requirements of a platform are used to generate
[...]

PR #1160, opened 2024-12-17
Workspaces Part I

rydrman:
This PR is intended to be the first of a few related to #1121 

This first PR introduces the `spk-workspace` crate that defines the structure of a `workspace.spk.yaml` file and updates the various spk commands to recognize it. By default, a "virtual" workspace is created which collects all `*.spk.yaml` files in the current directory (this is in-line with previous behavior). The only new feature that becomes unlocked by this is that you can create a `workspace.spk.yaml` that collects files from additional or different paths. 

For example, this repository now has a workspace file that makes `spk build spk-convert-pip` work from anywhere within it, finding and building from the spec file under the packages folder.


crates/spk-workspace/src/file_test.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[rstest]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir")]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir")]
```
jrray, line 62: Is the repetition here intentional?

jrray (changes requested):

crates/spk-workspace/src/file_test.rs
rydrman: ooops, ya probably not. I think I copy and pasted but never thought up another failure case to check...

crates/spk-workspace/src/file.rs
```diff
@@ -0,0 +1,151 @@
...
+        WorkspaceFile::from_yaml(workspace_file).map_err(LoadWorkspaceFileError::InvalidYaml)
+    }
+
+    /// Load the workspace for a given dir, looking at parent directories
+    /// as necessary to find the workspace root
+    pub fn discover<P: AsRef<Path>>(cwd: P) -> Result<Self, LoadWorkspaceFileError> {
+        let cwd = if cwd.as_ref().is_absolute() {
+            cwd.as_ref().to_owned()
+        } else {
+            match std::env::var("PWD").ok() {
```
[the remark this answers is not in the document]
rydrman: ya, I can - it's to avoid the expansion in the case of error messages, ya

crates/spk-workspace/src/file_test.rs
```diff
@@ -0,0 +1,77 @@
...
+#[rstest]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir")]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir/src/packages")]
+fn test_workspace_discovery(
```
jrray, line 63:
```suggestion
#[rstest]
#[case("", "")]
#[case("my-workspace", "my-workspace/src/packages")]
#[should_panic]
#[case("my-workspace", "other-dir")]
#[should_panic]
#[case("my-workspace", "other-dir/src/packages")]
fn test_workspace_discovery(
```

jrray (approved):

PR #1161, opened 2024-12-18
Workspaces Part II

rydrman:
Builds on #1160 as part of #1121

Makes use of the versions specified for any given template file by parsing the version from the request and only `finding` spec files that are appropriate. Uses the `LeastSpecifiedVersionRange` to allow partial versions to be given.

As a practical example, the `workspace.spk.yaml` from this repository lists versions for each python spec, and so you can run `spk build python/3` to build the default version of the python 3 package spec.


crates/spk-workspace/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+        source: spk_schema::Error,
+    },
+    #[error("Template cannot be loaded into workspace since it has no name defined: {file:?}")]
+    UnnamedTemplate { file: std::path::PathBuf },
+}
+
+#[derive(thiserror::Error, miette::Diagnostic, Debug)]
+pub enum LoadWorkspaceFileError {
+    #[error(
+        "workspace not found, no {} in {0:?} or any parent",
```
jrray, line 39: I remember seeing guidance somewhere that errors should start with lowercase letters (so they compose nicely), here there is a mix. I would say we generally start our errors with uppercase, but could get on board with switching to lowercase. Either way we should pick one and stick with it as best we can.

crates/spk-workspace/src/file.rs
```diff
@@ -0,0 +1,212 @@
...
+        deserializer.deserialize_any(RecipeCollectorVisitor)
+    }
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Ord, PartialOrd, Default)]
+pub struct TemplateConfig {
+    /// Ordered set of versions that this template can produce.
+    ///
+    /// An empty set of versions does not mean that no versions can
+    /// be produces, but rather that any can be attempted. It's also
```
jrray, line 133:
```suggestion
    /// be produced, but rather that any can be attempted. It's also
```

crates/spk-workspace/src/file_test.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[rstest]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir")]
+#[case("", "")]
+#[case("my-workspace", "my-workspace/src/packages")]
+#[should_panic]
+#[case("my-workspace", "other-dir")]
```
jrray, line 62: Duplicate cases again.

crates/spk-workspace/src/workspace.rs
```diff
@@ -0,0 +1,260 @@
...
+            by_name.push(ConfiguredTemplate { template, config });
+            Ok(by_name.last_mut().expect("just pushed something"))
+        }
+    }
+}
+
+/// The result of the [`Workspace::find_package_template`] function.
+// We are okay with the large variant here because it's specifically
+// used as the positive result of the function, with the others simply
+// denoting unique error cases.
```
jrray, line 203: Wouldn't a `Result` type with a tailored `Error` enum fit this situation better?

jrray (changes requested):

crates/spk-workspace/src/error.rs
rydrman: ya I am bad for this, and I think it's because of python usually using uppercase - though maybe that's not even true. I think I agree that lowercase makes sense as a practice

crates/spk-workspace/src/builder.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        Ok(self)
+    }
+
+    /// Add all recipe files matching a glob pattern to the workspace.
+    ///
+    /// If the provided pattern is relative, it will be relative to the
+    /// current working directory. All configuration for this path will be
+    /// left as the defaults unless already set.
```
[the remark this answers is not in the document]
rydrman: Ya, I think this is something that evolved over time and didn't get updated - it is based on the root if there is one

crates/spk-workspace/src/workspace.rs
rydrman: I tried this and the only challenge is that we cannot have `impl` for `Result` types and this type has functions on it - this is largely copied from the cli area so I didn't bother pushing too hard on a refactor in this change

Re: this comment, I actually managed to remove the lint error by returning references instead of owned values - so I will remove this block as well

crates/spk-workspace/src/lib.rs
```diff
@@ -2,6 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/spkenv/spk
 
+//! SPK workspaces are used to build a network of packages together.
+//!
+//! The [`WorkspaceFile`] is used to load [`Workspace`] configurations from
+//! yaml files on disk.
+
+#![deny(missing_docs)]
```
jrray, line 10: Thank you! There was more missing documentation than I realized when asking for this.

jrray (approved):

crates/spk-workspace/src/lib.rs
rydrman: ya, there's some that felt a little redundant but 🤷🏻

Issue #968, opened 2024-02-03
[BUG] `OverlayFSWithRenders` can fail abrupty on newer versions of linux


rydrman:
**Describe the bug**
Essentially, when running the tests the remount process eventually fails and the tests cannot continue. The output from mount in this case is not helpful:

```
/spfs: wrong fs type, bad option, bad superblock on none, missing codepage or helper program, or other error.
```

After some digging with strace, I found this culprit:

```
fsconfig(
  4, 
  FSCONFIG_SET_STRING, 
  "lowerdir", 
  "/tmp/spfs-runtime/lower:/tmp/spk-test-spfs-reporWeqZs/repo/renders/rbottriell/NX/OLKD6FCKQUZWSXINUOQ6PHMV26WE67S36ZCCXXMWLUD2WYQ6DQ====:/tmp/spk-test-spfs-reporWeqZs/repo/renders/rbottriell/AT/2LFH4PDZSRU422OUMNPN3EVF2GT2IJLVOWCV76TTLI2DBMQGAA====:/tmp/spk"..., 
  0
) = -1 EINVAL (Invalid argument)
```

Which lead me to this description of the same issue. Apparently, after this `fsconfig` logic was added, it reduced the effective limit of the overlayfs lowerdirs to 256 characters 😮 

https://github.com/util-linux/util-linux/issues/2287

For now, everything is fine when using `OverlayFsWithFuse`. Though it seems like we will need a way to detect if `fsconfig` will be used and further limit the length of the args...

davvid:
Sorry for not finding this bug earlier.

In case it's not obvious for others, configuring the following in spfs.toml avoids the errors for me as well:

```
[filesystem]
backend = "OverlayFsWithFuse"
```

Longer-term, do you think it'd be worth implementing the mount API calls directly so that we can follow the advice given in this commit message? https://github.com/brauner/linux/commit/aff097911207b1b941b0f1f22901b670cd3f0dc6

If we do that then we'd have to move away from the `mount` binary and issue the `fsconfig` calls directly. It's kinda unfortunate that the `mount` binary doesn't seem to have a way to handle this scenario so we'd have to basically implement this subset of mount ourselves. If you think that's worth it I might scratch away at it now that I know what's going on.

jrray:
[...]

PR #1164, opened 2024-12-28
env: fix handling of long lowerdir paths in newer kernel/mount versions

davvid:
The new "lowerdir+" option can be used to specify multiple layers when mounting. This avoids a limitation with the "lowerdir" option which is limited to a max of 256 characters.

Use the older options when the "legacy-mount-options" feature is enabled.

Fixes: #968

### DIFF
--- a/crates/spfs/Cargo.toml
+++ b/crates/spfs/Cargo.toml
@@ -18,6 +18,11 @@ default = []
 # If enabled, will create the "local" repository in a subdirectory
 # of the standard storage root, named "ci/pipeline_${CI_PIPELINE_ID}".
 gitlab-ci-local-repo-isolation = []
+# Use "mount" commands that are compatible with older centos7-era kernels that
+# do not support the "lowerdir+=" overlayfs options. "legacy-mount-options"
+# can run into path length limits when mounting many layers.
+# https://github.com/spkenv/spk/issues/968
+legacy-mount-options = []
 sentry = ["dep:sentry"]
 server = ["hyper/server", "tokio-util/codec", "tokio-util/io-util"]
 "protobuf-src" = ["dep:protobuf-src"]
--- a/crates/spfs/src/env.rs
+++ b/crates/spfs/src/env.rs
@@ -1126,10 +1126,19 @@ pub(crate) fn get_overlay_args<P: AsRef<Path>>(
     // the rightmost on the command line is the bottom layer, and the
     // leftmost is on the top). For more details see:
     // https://docs.kernel.org/filesystems/overlayfs.html#multiple-lower-layers
-    args.push_str("lowerdir=");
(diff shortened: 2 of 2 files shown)


rydrman (approved):
Thanks for this! I agree that it's an improvement regardless of the other related discussions on the issue.

davvid:
Thanks for reviewing. If you're comfortable merging now feel free. Otherwise, I'll wait until next week before merging in case anyone brings up any backwards-compat concerns.

I think we've all either left centos7-era distros behind or are using fuse so it doesn't seem like this should be an issue, tho.

rydrman:
Thanks, I think I am waiting on confirmation from @jrray that they are ready/okay with this at SPI

jrray:
Sorry for the delay responding to this.

I think to no one's suprise, this change errors out on CentOS 7 when trying to mount /spfs:

    mount: wrong fs type, bad option, bad superblock on none,
           missing codepage or helper program, or other error
    
           In some cases useful info is found in syslog - try
           dmesg | tail or so.
    Error: spfs::generic (link)
    
      × Failed to mount overlayfs
    
    [13065739.400610] overlayfs: unrecognized mount option "lowerdir+=/tmp/spfs-runtime/lower" or missing value

I get a similar result on Ubuntu 22.04.

This change will break things for us and it needs to be gated behind a feature flag, a configuration option, or some kind of runtime detection of supported kernel versions.

davvid:
Thanks for verifying. Any thoughts on what we should name the feature flag? My sug would be `mount-v1` and it would let older systems opt in to the original implementation.

In other words, this would be the new default behavior (favoring newer systems) and builds for older kernels can opt-out of the new behavior by enabling the `mount-v1` feature.

If that sounds good I'll update this to take that into account.

jrray:
I'm okay with that approach. I think we have some other flags using "legacy" and I could imagine this one being something like "legacy-mount-options". It's the "v1" that implies we should have a "v2" too and so on.

The other thought is if we know the kernel version where this change starts being supported then that could be incorporated into the name somehow. "kernels-before-xxx"

davvid:
I rebased this change and updated it to retain the original behavior when the "legacy-mount-options" feature is enabled.

jrray:
Thanks for adding the feature. LGTM!

jrray:
Circling back on this to mention that the new `+=` syntax is not supported on Rocky 9.4.

davvid:
That's what I get for being on a rolling release distro ;-). I'll try and test RHEL/Rocky 9.6 soon to get a better gauge on whether it's supported there or not.

I don't mind flipping the feature flag around to make this opt-in rather than opt-out. I'd basically rename the feature to be `new-mount-options` instead of `legacy-mount-options`.

PR #1168, opened 2025-01-14
Aligning spk-convert-pip with current SPI changes

dcookspi:
This modifies the license field processing in `spk-convert-pip` to align it with the current SPI license processing. The changes are from last September and move a code block around, as well as removing the "Unknown" default value.

This also versions up the `spk-convert-pip` spk package to `v1.4.4` to sync up the version numbers

### DIFF
--- a/packages/spk-convert-pip/spk-convert-pip
+++ b/packages/spk-convert-pip/spk-convert-pip
@@ -291,23 +291,11 @@ class PipImporter:
         assert not info.requires_external, "No support for external requirements"
         assert not info.supported_platforms, "No support for supported platforms field"
 
-        # The spk default is "Unlicensed", which seems odd
-        package_license = "Unknown"
-        if hasattr(info, "license"):
-            if info.license is not None:
-                package_license = str(info.license)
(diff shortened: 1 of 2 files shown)


packages/spk-convert-pip/spk-convert-pip
```diff
@@ -291,23 +291,11 @@ class PipImporter:
...
-                        package_license[:LICENSE_FIELD_TRUNCATION_POINT]
-                        + TRUNCATED_VALUE_INDICATOR
-                    )
-
         spec = {
             "pkg": f"{_to_spk_name(info.name)}/{_to_spk_version(info.version)}",
             "api": "v0/package",
             "sources": [],
             "meta": {
-                "license": package_license,
```
rydrman, line 310: still put this in the metadata, but you _should_ be able to disable the license check - maybe this needs to be split out first to make this possible.

Aka split SpdxLicense rule into two (has license vs is spdx) and then make the semantics such that both can be disables in these cases so that the acutal included license is never checked.

```
build:
  validation:
    - rules: 
        - allow: SpdxLicense
        - allow: HasLicense
```

rydrman (approved):

PR #1173, opened 2025-01-29
Bugfix: Solver building from source package with recursive dependencies

dcookspi:
This fixes an issue in the Solver: when `--allow-builds` is enabled for the Solver (so, binary only is set to false and the `BinaryOnlyValidator` is not enabled), and it tries to build a package from source, and that package contains a recursive build dependency, which can generate a recurisve loop where the Solver tries to continually the same package build from source.

Adds a `new_builds_started` field to the Solver. When the Solver is allowed to build from source, it uses that new field to detect recursive loops in build environment dependencies.

Found this because we have a converted `python-pip` package with a build dependency on `python-pip` that's part of its manual bootstrapping.  When an `--allow-builds` solve contained a request for it, and the first version the solver found didn't have a build for the current OS, it triggered this issue.

### DIFF
--- a/crates/spk-solve/src/solver.rs
+++ b/crates/spk-solve/src/solver.rs
@@ -20,7 +20,16 @@ use spk_schema::foundation::version::Compatibility;
 use spk_schema::ident::{PkgRequest, Request, RequestedBy, Satisfy, VarRequest};
 use spk_schema::ident_build::EmbeddedSource;
 use spk_schema::version::{ComponentsMissingProblem, IncompatibleReason, IsSameReasonAs};
-use spk_schema::{try_recipe, BuildIdent, Deprecate, Package, Recipe, Spec, SpecRecipe};
+use spk_schema::{
+    try_recipe,
+    BuildIdent,
+    Deprecate,
+    Package,
+    Recipe,
+    Spec,
+    SpecRecipe,
+    VersionIdent,
+};
 use spk_solve_graph::{
     Change,
     Decision,
@@ -123,6 +132,9 @@ pub struct Solver {
     // highlight problem areas in a solve and help user home in on
     // what might be causing issues.
     problem_packages: HashMap<String, u64>,
+    // Set of package/versions the solver has decided to try to build
+    // from source as part of a solve
+    new_builds_started: HashSet<VersionIdent>,
 }
 
 impl Default for Solver {
@@ -141,6 +153,7 @@ impl Default for Solver {
             number_of_steps_back: Arc::new(AtomicU64::new(0)),
             error_frequency: HashMap::new(),
             problem_packages: HashMap::new(),
+            new_builds_started: HashSet::new(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/spk-solve/src/solver.rs
```diff
@@ -1106,6 +1135,12 @@ impl Solver {
         }
     }
 
+    /// Add the given set of package/versions to the solver's new
+    /// builds started records
+    pub fn add_new_builds_started(&mut self, new_builds: &HashSet<VersionIdent>) {
```
jrray, line 1140: Rust design nit: if a method needs an owned value, use an owned value for the parameter. Let the caller decide if it needs to clone the value.

crates/spk-solve/src/solver.rs
```diff
@@ -349,7 +362,17 @@ impl Solver {
...
     /// spec.
     #[async_recursion::async_recursion]
-    async fn resolve_new_build(&self, recipe: &SpecRecipe, state: &State) -> Result<Arc<Spec>> {
+    async fn resolve_new_build(&mut self, recipe: &SpecRecipe, state: &State) -> Result<Arc<Spec>> {
+        // Check if the package/version is one that the solver has
+        // already started to resolve a new build for, and error out
+        // if it is. The new_build_started set's insert will return
+        // true if the set did not contain the value before the insert.
+        if !self.new_builds_started.insert(recipe.ident().clone()) {
+            return Err(Error::String(
```
jrray, line 371: I'm expecting to see a custom error type for this situation and something that handles this situation.

For example, loops when exploring the build-from-src option aren't necessarily hard errors because there could be alternative builds available.

What currently happens when this error is returned?

jrray (changes requested):
Changes are looking good but really hoping to see a test case added to exercise this situation.

crates/spk-solve/src/solver.rs
dcookspi: It gets called like this from the solver's step_state():
```
let new_spec = match self.resolve_new_build(&recipe, &node.state).await {
    Err(err) => {
        notes.push(Note::SkipPackageNote(
            SkipPackageNote::new_from_message(
                spec.ident().to_any_ident(),
                format!("cannot resolve build env for source build: {err}"),
            ),
        ));
        self.number_builds_skipped += 1;
        continue;
    }
    res => res?,
};
```
I'd initially thought it would expect and handle a compat result, but it handles any error by treating it a build that can't be used.  

I'll add custom error type for this situation though.

dcookspi:
> Changes are looking good but really hoping to see a test case added to exercise this situation.

- [ ] add a test

crates/spk-solve/src/solver.rs
dcookspi: I've swapped this around.

crates/spk-solve/src/error.rs
```diff
@@ -65,6 +66,10 @@ pub enum Error {
     SolverLogFileIOError(#[source] std::io::Error, PathBuf),
     #[error("Error: Flushing solver log file: {0}")]
     SolverLogFileFlushError(#[source] std::io::Error),
+    #[error(
+        "Cannot build package ({0}) from source during a solve when it has a dependency on itself"
+    )]
+    SolverBuildFromSourceDependencyLoopError(VersionIdent),
```
rydrman, line 72: This feels like it would be worth having a documented error code for since additional explanation and context and help might be useful
```suggestion
    #[diagnostic(
        code(spk::solve::recursive_build_dependency),
        help("this package requires an older version of itself to build but none exists")
    )
    SolverBuildFromSourceDependencyLoopError(VersionIdent),
```

crates/spk-solve/src/error.rs
```diff
@@ -65,6 +66,10 @@ pub enum Error {
     SolverLogFileIOError(#[source] std::io::Error, PathBuf),
     #[error("Error: Flushing solver log file: {0}")]
     SolverLogFileFlushError(#[source] std::io::Error),
+    #[error(
+        "Cannot build package ({0}) from source during a solve when it has a dependency on itself"
```
rydrman, line 70:
```suggestion
        "Cannot build package ({0}) from source during a solve because it has a dependency on itself"
```

just want to make sure that this is not always an error - eg if an older version did exist you would be able to form the build environment and proceed with this solve, right?

rydrman (commented):
just a couple of notes on documentation

crates/spk-solve/src/error.rs
dcookspi: I've updated this.

crates/spk-solve/src/error.rs
dcookspi: I've updated this.

dcookspi:
Did the remaining updates, rebased, and fixed test and new method to account for the Solver/SolverImpl changes made since this was originally done.

PR #1176, opened 2025-02-04
Add `--prune-repeated-keep <number>` option to `spfs clean` 

dcookspi:
This adds a `--prune-repeated-keep <number>` option to `spfs clean`. It allows the user to decide how many repeated tags in a tag stream to keep. The existing `--prune-repeated` option is equivalent to `--prune-repeated-keep 1`. But either option can be used. 

We have a case where we'd like to keep some repeated tags, say the last 20 , and prune the rest periodically from the local repo.


dcookspi:
I left a default value in, I'll need to fix that before this goes further.

- [ ] Remove the default value and use NonZero

rydrman (approved):

PR #1180, opened 2025-02-13
Bugfix: host options were being left out of the solver's initial requests

dcookspi:
This fixes a bug where the host options were being left out of the solver's initial requests. A few lines of code hand been mistakenly removed in an earlier change. This restores them and adds a test for this situation.

### DIFF
--- a/crates/spk-cli/common/src/flags_test.rs
+++ b/crates/spk-cli/common/src/flags_test.rs
@@ -36,14 +36,17 @@ fn test_option_flags_parsing(#[case] args: &[&str], #[case] expected: &[(&str, &
     assert_eq!(actual, expected);
 }
 
+#[rstest]
+#[case::no_host_true(true)]
+#[case::no_host_false(false)]
 #[tokio::test]
-async fn test_get_solver_with_host_options() {
+async fn test_get_solver_with_host_options(#[case] no_host: bool) {
     // Test the get_solver() method adds the host options to the solver
     // correctly.
 
     let options_flags = crate::flags::Options {
         options: Vec::new(),
-        no_host: false,
+        no_host,
     };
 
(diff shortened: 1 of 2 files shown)


crates/spk-cli/common/src/flags_test.rs
```diff
@@ -33,3 +35,38 @@ fn test_option_flags_parsing(#[case] args: &[&str], #[case] expected: &[(&str, &
...
 }
+
+#[tokio::test]
+async fn test_get_solver_with_host_options() {
+    // Test the get_solver() method adds the host options to the solver
+    // correctly.
+
+    let options_flags = crate::flags::Options {
+        options: Vec::new(),
+        no_host: false,
```
jrray, line 46: I suggest using rstest and creating test cases for both true and false here, to compliment this test.

crates/spk-cli/common/src/flags_test.rs
```diff
@@ -33,3 +35,38 @@ fn test_option_flags_parsing(#[case] args: &[&str], #[case] expected: &[(&str, &
...
+        check_impossible_initial: false,
+        check_impossible_validation: false,
+        check_impossible_builds: false,
+        check_impossible_all: false,
+    };
+
+    let solver = solver_flags.get_solver(&options_flags).await.unwrap();
+    let initial_state = solver.get_initial_state();
+
+    for (name, value) in HOST_OPTIONS.get().unwrap() {
```
jrray, line 68: There should be an assertion that  `TEST_OPTIONS.get()` is not empty, so we know this is really testing something. If empty this test would have no assertions.

jrray:
David is out today so I addressed my own notes to move this forward.

rydrman (approved):

PR #1185, opened 2025-03-04
Fix runtime flattened_layers getting lost

jrray:
When `skip_runtime_save` was added, when true it unintentionally did not update the `flattened_layers` field of the runtime. This then made flattened layers subject to premature deletion by spfs-clean. Deleting in-use renders can cause existing runtimes to be left with an empty /spfs directory.

### DIFF
--- a/crates/spfs/src/resolve.rs
+++ b/crates/spfs/src/resolve.rs
@@ -340,11 +340,11 @@ where
 
     // Note the layers we manufactured here via flattening so they will have a
     // strong reference in the runtime.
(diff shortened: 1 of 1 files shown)


rydrman (approved):

rydrman:
Might be good to add a test case if it's not difficult

jrray:
> Might be good to add a test case if it's not difficult

I wanted to add a test, but believe the code needs to be restructured to create a meaningful test. Meanwhile this is a critical bug for us so I will merge this as is and circle back.

PR #1187, opened 2025-03-07
Add Platform V1 Spec

rydrman:
The next stage in work for #1121 which adds a new way to specify platforms with static build version as well as separated build and runtime requests.

Here is the updated doc page for reference, note that there is no information about the static build version yet as it's not used by anything:

---
# Specifying Platforms

As a concept, platforms are packages that only provide _opinions_ on what should exist in an environment. They contain no files or code, and have no concrete dependencies but limit what versions are allowed in order to enforce consistency across teams, applications and developers.

As an example, the [VFX Reference Platform](https://vfxplatform.com/) provides "a set of tools and library version to be used as a common target platform for building software for the VFX industry". Platforms in SPK were designed to be able to describe and enforce these kind of recommendations.

### Platforms are Packages

At their core, platforms are a convent way to write package specs that only provide an opinion on what versions can be used and nothing else. Once defines, platforms are still built, published, and used the same way as any other package.

A platform spec reduces the amount of boilerplate needed to set up a platform package compared to using the full `v0/package` recipe format. A platform spec will be filled in with appropriate defaults automatically for a platform-style package.

### Creating Platforms

Below is an example platform

```yaml
api: v1/platform
platform: company-platform/2025.01.01
requirements:
  - pkg: gcc
    atBuild: =9.3.1
    atRuntime: 9.3.1
  - pkg: python
    atBuild: =3.9.18
    atRuntime: 3.9
  - pkg: imath
    atBuild: =3.0.1
    atRuntime: 3
```

The `platform:` fields provides the name of the platform package. The
`api:` field indicates this is a platform spec.

The `requirements` field contains the list of requirements in the
platform. In each case, the requirement specifies the version request to be
used at build time vs at run time. This allows the platform to differently
constrain the requirement depending on if it's being used to define a build
environment or a runtime environment.

These requirements are placed onto the `build` and `run` components of the package
generated by this platform, respectively. The generated requirements will have
`include: IfAlreadyPresent` added to them automatically, it does not need to be
specified for them.

#### Inheritance

Additionally, the platform spec provides a way to inherit requirements from another package. Typically, it makes the most sense to inherit from another platform, but it could be any package.

By inheriting from base packages, platforms can augment and override their requirements. For example, a DCC (Digital Concent Creation) application platform can inherit the requirements from company or site-specific platform and then introduce additional constraints for the application itself.

Here is an example platform spec that inherits from the 'company-platform' and makes adjustments of its own:

```yaml
api: v1/platform
platform: dcc-platform/2025.01.01
base:
  - company-platform/2024.01.01
requirements:
  # this DCC uses an older version of python
  - pkg: python
    atBuild: =3.7.10
    atRuntime: 3.7
  # it also has no need to constrain the version of imath that's used
  - pkg: imath
    atBuild: false
    atRuntime: false
```

The `base:` field indicates which platform this one inherits from. In the case of multiple bases, each one simply overrides the previous in turn.

Any `atBuild` or `atRuntime` value that is specified in the platform spec will override the same value for the named requirement in the bases. Conversely, this also means that if a requirement omits either of the `atBuild` or `atRuntime` fields, then the value from the base will still remain.

The special `false` value can be used to remove the requirement entirely.

This is the complete specification of the same resulting `dcc-platform` without basing it on the `company-platform`:

```yaml
api: v1/platform
platform: dcc-platform
requirements:
  - pkg: gcc
    atBuild: =9.3.1
    atRuntime: 9.3.1
  - pkg: python
    atBuild: =3.7.10
    atRuntime: 3.7
  # - pkg: imath not present
```


rydrman:
From the meeting today:
- reference docs for v0 platform and v1 platform (like we do for spec)

crates/spk-schema/src/requirements_list.rs
```diff
@@ -24,30 +26,56 @@ mod requirements_list_test;
...
+where
+    R: Named<OptName>,
+{
+    /// Get a reference to the first request with the given name
+    pub fn get(&self, name: impl AsRef<str>) -> Option<&R> {
+        self.0.iter().find(|r| r.name() == name.as_ref())
+    }
+
+    /// Get a mutable reference to the first request with the given name
+    pub fn get_mut(&mut self, name: impl AsRef<str>) -> Option<&mut R> {
```
jrray, line 69: This allows the caller to violate the unique name invariant by changing the name of item returned.

crates/spk-schema/src/requirements_list.rs
```diff
@@ -58,6 +86,19 @@ impl RequirementsList {
         None
     }
 
+    /// Remove a requirement from this list.
+    ///
+    /// All requests with the same name as the given name are removed
+    /// from the list.
```
jrray, line 92: No arguments about the implementation here but there is an invariant that names can't be repeated, so the way this is worded is a bit misleading since it implies it could be normal for there to be multiples.

I recognize that this is an existing docstring that is just being relocated.

crates/spk-schema/src/v1/platform_test.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    let spec: Platform = serde_yaml::from_str(
+        r#"
+         platform: test-platform
+         base: [base]
+         requirements:
+           - pkg: inherit-me1
+             atRuntime: 2.0.0
+           - pkg: inherit-me2
+             atRuntime: false
```
jrray, line 295:
```
           - pkg: inherit-me2
             atBuild: =1.0.0
             atRuntime: false
```
What would happen if I wrote this?

Will it be legal to say a package is required [to be a specific version] for build environments but have no influence on runtime environments?

I'm not necessarily arguing for that to be a feature, but looking at this test behavior it is treating this as deleting `inherit-me2` from the list. Would the same happen if this said `atBuild: false` instead of `atRuntime: false`?

If it is not possible to mix and match false and not-false then I would suggest a different way of expressing deletes, like we have in v0.

jrray (changes requested):

docs/use/spec.md
```diff
@@ -414,85 +418,6 @@ install:
...
-    - pkg: python: 3.7
-  remove:
-    - pkg: imath
-```
-
-The `base:` field indicates which platform this platform spec is based on (inherits the requirements from).
-
-Specifying a requirement directly with `- ` is the same as specifying it with `add:`, but is a shorthand for convenience.
-
-The `add:` and `remove:` entries indicate changes to the requirements inherited from the base platform. `add:` means "add or replace". `remove:` means remove entirely. `remove:` will work on components of packages, if they are specified, as well as full packages. Removing requirement is done before adding when determining the platform's final requirements.
```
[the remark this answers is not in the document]
rydrman: yes, that's why the `atBuild` and `atRuntime` allow the special `false` value

crates/spk-schema/src/v1/platform_test.rs
rydrman: yes, this is possible and does as you describe. The build and runtime are treated entirely separately so it's legal. In this case it's only being deleted `atRuntime` but the base also had no requirement on `atBuild` so the result is no restrictions on either.

crates/spk-schema/src/requirements_list.rs
rydrman: ya, I thought about this as well. I decided it was better to be explicit about what it actually did but just in case it tricked someone later... I think I'll just add even more detail here to call that invariant out

crates/spk-schema/src/requirements_list.rs
rydrman: ya, I think I added this for completeness but you're right - it's also not used AFAICT

crates/spk-schema/crates/ident/src/request.rs
```diff
@@ -183,14 +183,20 @@ pub enum Request {
...
             Request::Pkg(r) => r.pkg.name.as_opt_name(),
         }
     }
+}
+
+impl Request {
+    /// Return the canonical name of this request.
+    pub fn name(&self) -> &OptName {
+        spk_schema_foundation::spec_ops::Named::name(self)
+    }
```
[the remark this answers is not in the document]
rydrman: nope, I'll remove it

crates/spk-schema/crates/foundation/src/version_range/mod.rs
```diff
@@ -1465,6 +1465,25 @@ impl FromStr for VersionFilter {
...
+        Self::from_str(&raw).map_err(serde::de::Error::custom)
+    }
+}
+
+impl serde::ser::Serialize for VersionFilter {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
```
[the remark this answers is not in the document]
rydrman: ya, I don't disagree - probably doesn't matter for now but something we could reconsider. I think the recursiveness was more to fix issues with interactions via the trait than any actual requirement of behaviour.

rydrman:
I added the additional reference docs for the old and new platform types. They are fairly bare bones at the moment but we can iterate. I've reorganized the reference pages by api version and name now, with an alias to redirect any old links to the v0/package reference.

docs/ref/api/v1/platform.md
```diff
@@ -0,0 +1,31 @@
...
+| compat       | _[Compat]({{< ref "../v0/package#compat" >}})_               | The compatibility semantics of this packages versioning scheme                                                                                        |
+| deprecated   | _boolean_                                                    | True if this package has been deprecated, this is usually reserved for internal use only and should not generally be specified directly in spec files |
+| base         | _List[[Identifier]({{< ref "../v0/package#identifier" >}})]_ | (Optional) Base packages to inherit requirements from                                                                                                 |
+| requirements | _List[[Requirement](#requirement)]_                          | The set of requirements for this platform                                                                                                             |
+
+## Requirement
+
+Each platform requirement names a package and the constraints for that package in downstream environments.
+
+Like an [Identifier](#identifier) but with a version range rather than an exact version, see [versioning]().
```
jrray, line 25: Is the empty url here intentional? What does it do?

jrray (approved):
Approving but unsure about that empty url.

docs/ref/api/v1/platform.md
rydrman: not intentional, I'll put the right ref in there, thanks

PR #1189, opened 2025-03-12
Updated ring dependency to 0.7.14 to avoid a bug in 0.7.13

dcookspi:
This updates the `ring` dependency to `0.7.14` to avoid a bug in `0.7.13` that impacts centos 7 compiles.


rydrman (approved):

jrray:
FYI this PR modified Cargo.toml but did not commit the corresponding change to Cargo.lock, so all PRs against main for the moment will contain the Cargo.lock change.

PR #1190, opened 2025-03-13
Solve `spk info filename.spk.yaml` failing with file not found

jrray:
At least for the simple case where no workspace file is present, fix `find_package_template_by_file` so it compares with canonicalized path names.

Change `Workspace::load_or_default` so that if the `--workspace` argument is a directory (that doesn't contain a workspace.spk.yaml file) it still roots the virtual workspace to the given directory. This allows for workspace tests to load a virtual workspace in a temporary directory.

Add a very basic cmd_view test that verifies it is possible to use `spk view --variants filename.spk.yaml`, something which is required by the CI automation at SPI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4083,6 +4083,7 @@ dependencies = [
  "futures",
  "itertools 0.12.1",
  "miette",
+ "rstest 0.25.0",
  "serde",
  "serde_json",
  "serde_yaml 0.9.27",
@@ -4093,8 +4094,10 @@ dependencies = [
  "spk-storage",
  "spk-workspace",
  "strum",
+ "tempfile",
  "tokio",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -4660,6 +4663,7 @@ name = "spk-workspace"
 version = "0.42.0"
 dependencies = [
  "bracoxide",
+ "dunce",
  "format_serde_error",
  "glob",
  "itertools 0.12.1",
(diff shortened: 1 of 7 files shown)


rydrman:
There's a larger question here of what we want to support with respect to workspaces. I think that these tests are failing appropriately (because there is a defined workspace with nothing in it). Albeit a better error would be good, but I think the more relevant scenario is referencing files not covered by the default pattern in a virtual workspace - for which we may want to discuss what's the most approprate.

Just to check your temperature on this, though - would you be okay with a warning in these cases? Eg we load the file if possible but say that it's not within the current workspace?

jrray:
You're referring to the tests in the other PR? My position is all those tests should pass:

1. workspaces should be an opt-in feature however certain commands now always operate on workspaces even if no workspace.spk.yaml file exists
2. `spk info <valid recipe filename>` should be able to tell you something useful about the recipe regardless of any workspace configuration
3. In particular, `spk info --variants <valid recipe filename>` _must_ be able to report on the variants defined in a recipe regardless of any workspace configuration since this is an important part of our CI automation

Only a couple of the new tests include creating a workspace.spk.yaml file, the rest are showing misbehavior when no workspace exists.

`spk build <valid recipe filename>` works with or without a workspace.spk.yaml file that does or doesn't include said recipe. I think this is a good thing and is the proper behavior. I could be convinced that some operations might warn in the presence of a workspace but I do not agree with breaking existing behavior when no workspace.spk.yaml file exists.

The moment I installed the new spk build on our CI runners, _every_ CI job started to fail.

rydrman:
Okay, ya I think we are aligned on all of this - I'm addressing this today, and think the ideal solution is something that makes the use of the workspace on the command line more explicit... TBD

rydrman (approved):

PR #1191, opened 2025-03-13
Fix workspace issues stemming from new virtual workspace no accounting for path arguments

jrray:
The new workspace feature has introduced a variety of regressions for
using `spk info --variants <filename>` and will error with a file not
found error for files that do actually exist.

One problem is that when a workspace spec doesn't exist, then the only
possible files that can be inspected have to exist in the root of the
virtual workspace generated, ignoring that the command line contains the
name of a valid file.

Another problem is when the workspace does exist, then the only possible
files that can be inspected have to be defined in the workspace, even
when the command line contains the name of a valid file.

Perhaps the design can be reconsidered to not use a "virtual workspace"
in these cases or at least have a fallback that tries to open the file
directly.

### DIFF
--- a/crates/spk-workspace/src/workspace.rs
+++ b/crates/spk-workspace/src/workspace.rs
@@ -9,7 +9,7 @@ use spk_schema::name::{PkgName, PkgNameBuf};
 use spk_schema::version_range::{LowestSpecifiedRange, Ranged};
 use spk_schema::{SpecTemplate, Template, TemplateExt};
 
-use crate::error;
+use crate::error::{self, BuildError};
 
 #[cfg(test)]
 #[path = "workspace_test.rs"]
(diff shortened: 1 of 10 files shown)


jrray:
@rydrman I took the liberty of assigning this to you for you to decide on a path forward.

jrray:
I just saw the build fails but the tests here can be run with `cargo test -p spk-cli-group4 --lib`. I took back the assignment and with fix the other build errors.

rydrman:
thanks, I'll take a look shortly

rydrman:
I ended up taking a more simplistic approach of allowing files to be loaded dynamically into workspaces. There might be other implications to this which we need to refine in the future, but it more accurately represents the previous behavior and allows things to move forward - let me know if that tracks for you

jrray (commented):
Sorry for the delay, I have been ill. LGTM but I am unable to approve because I authored some of this commits in this PR.

rydrman:
Okay, I've squashed a couple of commits together into something that makes sense to me as a flow of change.

jrray:
Thanks but it still considers my an author.

rydrman:
I think that's okay, since you did write the tests anyway - unless I'm missing something?

jrray:
I just mean I still can't approve the PR. I tagged @dcookspi as a reviewer.

crates/spk-workspace/src/workspace.rs
```diff
@@ -62,21 +62,47 @@ impl Workspace {
...
+        S: AsRef<str>,
+    {
+        if let Err(FindPackageTemplateError::NotFound(_)) =
+            self.find_package_template(package.as_ref())
+        {
+            if std::fs::exists(package.as_ref()).ok().unwrap_or_default() {
+                self.load_template_file(package.as_ref())?;
+            }
+        }
+        // NOTE: it would be preferable not to run this function twice and instead return
```
jrray, line 94: This could be a situation where the use of unsafe is justified.

PR #1195, opened 2025-03-21
Skip invalid template files when loading workspace

jrray:
For the use case of `spk info --variants <valid spec file>`, this command would fail with an error about a different file if another file exists with a name matching `*.spk.yaml` but loading this other file fails for any reason.

Like #1190 and #1191, this command is expected to succeed and is not intended to interact with any other files other than the one named on the command line.

### DIFF
--- a/crates/spk-cli/common/src/flags.rs
+++ b/crates/spk-cli/common/src/flags.rs
@@ -723,6 +723,7 @@ impl Workspace {
                     .with_glob_pattern("*.spk.yaml")?
                     .build()
                     .into_diagnostic()
+                    .wrap_err("loading *.spk.yaml")
             }
             Err(err) => Err(err.into()),
         }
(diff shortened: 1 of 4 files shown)


jrray:
This represents another failure mode we encountered at SPI. Someone was using a file with the `.spk.yaml` extension for a spawn target spec and now it was impossible to use `spk info --variants <filename>` for any valid spk recipe file in that directory.

Skipping invalid files seems like a good course of action, however if there is a use case for making this kind of thing a hard error then it would be best to add a test that demonstrates that use case. We can make warning vs erroring a runtime decision based on context.

An unintended consequence of this change is that now if the doing `spk info --variants demo.spk.yaml` and `demo.spk.yaml` has some problem that makes it fail to load, we'll first say we didn't find that recipe template:

image

This is a little better than literally saying "file not found" but perhaps it could say something like "did not find a valid recipe template" since the rest of the error goes on to say more specific things about a file.

rydrman (approved):

PR #1203, opened 2025-04-24
Dependency Upgrades

rydrman:
Each commit handles one aspect of this upgrade, with hyper and the web stack being the most involved diff.

I didn't look at every dependency, but tried to focus on the larger crates or ones that ended up with multiple versions in the lock file because of diversions with transient dependencies from other crates.

I also did a little alphabetical sorting in the toml where I was touching things.


crates/spfs/src/storage/fs/renderer_unix.rs
```diff
@@ -470,7 +470,7 @@ where
                         committed_path.as_path(),
                         Some(target_dir_fd),
                         std::path::Path::new(entry.name()),
-                        nix::unistd::LinkatFlags::NoSymlinkFollow,
+                        nix::fcntl::AtFlags::AT_SYMLINK_FOLLOW,
```
rydrman, line 473: I'm not sure that this change will have any real-world effect, but I did start to see EINVAL with the previous argument (maybe because the source was not a symlink). In any case, I don't see a reason not to evaluate symlinks, despite there never being a likely scenario where there is one since the link will always still fail across devices.

jrray: There's no comment here giving insight to why it was the previous value, or now why it is the current value, so 🤷 .

jrray (approved):

PR #1209, opened 2025-05-12
Fix developer docs link

BryceGattis:
Change link to link to built documentation instead of GitHub src folder.


rydrman (commented):
I'm fine with this change, I agree that the public url is a better interface for people to read the docs in. For consistency, though, there are some other links on line 26 that I think should be updated to match

rydrman (approved):

PR #1210, opened 2025-05-12
Add docs build instructions

BryceGattis:
Update docs to work with Windows by relying on config change rather than symlinks. Also added build instructions to a new README.md


docs/README.md
```diff
@@ -0,0 +1,14 @@
+# Build instructions
+
+To build the docs you must install the following requirements:
+
+- [Hugo 0.147.1](https://github.com/gohugoio/hugo)
```
rydrman, line 5:
```suggestion
- [Hugo 0.147+](https://github.com/gohugoio/hugo)
```

I don't think we care about a particular release, since I've yet to see breaking changes come through hugo. This change feels like it gives some sense but doesn't tie the docs to a version that might get outdated

docs/README.md
```diff
@@ -0,0 +1,14 @@
+# Build instructions
```
rydrman, line 1:
```suggestion
# Build Instructions
```

docs/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+To build the docs you must install the following requirements:
+
+- [Hugo 0.147.1](https://github.com/gohugoio/hugo)
+
+To build the docs run the following:
```
rydrman, line 7:
```suggestion
To build the docs and serve them in a local hot-reload server for development, run the following:
```

rydrman (commented):
As long as the CI runs fine with the configuration change, I have just a couple updates for clarity

BryceGattis:
@rydrman Is it possible to access the docs that are built in this PR? I see there's a CI action to build it but I can't seem to access any hosted docs from it.

rydrman:
No, there's no staging site or anything - they have to be tested locally with hugo

BryceGattis:
@rydrman Is a staging site something we could add in our CI? If it only ran when doc src files changed it shouldn't run too often right?

rydrman:
Currently our site is generated separately from the gitlab ci process so AFAICT there's no easy way to turn on per-branch deployments. We talked about moving to github pages and if that would work but it's not currently on the table for getting done. Given that hugo is so easy to run locally, there's not a huge need - though it would be nice

PR #1211, opened 2025-05-12
Feature/publishing packages doc page

BryceGattis:
Add new publishing packages doc page since we didn't seem to have anything that explained what "publishing" a package meant and how to configure it.

Would be awesome if we could break out our configuration settings individually and could link to them like [this](https://rez.readthedocs.io/en/stable/configuring_rez.html#local_packages_path). That way this doc page could just link to these hyperlinks rather than duplicate the config info.

TODO:
- Need some help confirming exactly what `spk publish` does at the bottom of this new doc page.


rydrman:
just noting that we do have this tiny section, which should either link or is potentially a place to put additional detail: https://spkenv.dev/use/#publish-a-package

rydrman:
just realizing that this is a merge request not an issue (thanks!) so maybe a link from that area mentioned above would be prudent

BryceGattis:
Done. Just need more info on any additional processes that run when running `spk publish` if there are any.

PR #1212, opened 2025-05-13
Fix `spfs clean` failing if a tag namespace exists

jrray:
This slipped through testing, the clean would immediately fail when it encounters a directory named like "test#ns" as it fails to parse the tag spec, reporting that the "org" is invalid.

This error is avoided by allowing "#ns" to be the suffix of an org and then `spfs clean` seems to work as expected, but ignores all the cases where it _shouldn't_ accept an "org" that contains a namespace.

The current tag stream code doesn't really understand tag namespaces or offer a way to choose to walk into them or skip them. It was intended that `spfs clean` would visit all tags in all namespaces, but otherwise other tag operations would only see tags in the configured namespace.

### DIFF
--- a/crates/spfs-cli/cmd-clean/src/cmd_clean.rs
+++ b/crates/spfs-cli/cmd-clean/src/cmd_clean.rs
@@ -50,6 +50,11 @@ pub struct CmdClean {
     #[clap(long)]
     dry_run: bool,
 
+    /// Prune all tag namespaces. The default is to only prune tags in the
+    /// active tag namespace.
+    #[clap(long)]
+    prune_all_tag_namespaces: bool,
+
     /// Prune old tags that have the same target as a more recent version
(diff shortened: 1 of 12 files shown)


rydrman (approved):

rydrman:
From the meeting today:
  - ultimately, the repos should have an API that allow clients to list the available namespaces for the cleaner, or some additional extension trait that has a separate namespace-free iterator that also identifies the namespace for each of the yielded tags
  - we should consider how the pruning of tags plays with this
  - alternatively, in the interim it could skip files that create an invalid tag name... but this has the issue of the cleaner not seeing all objects and possibly cleaning too much
  - this test is valuable but the current fix might need to be revisited

jrray:
@rydrman there have been a number of commits here since you last looked/approved this.

rydrman (approved):

PR #1213, opened 2025-05-13
Clarify Windows build instructions

BryceGattis:
`choco` is not installed by default on Windows, so we need to install that first. I've clarified the README.md to try using Windows Package Manager `winget` to get Chocolatey first, then we can proceed as written.


rydrman (approved):
great addition, thanks for this!

BryceGattis:
@rydrman Looks like spell check is failing due to `winget`. I'm not sure how to fix this. Is there a list of ignored words or something?

rydrman:
Yes, we use cspell there's a config file in the root of the repo

Issue #1218, opened 2025-05-22
[BUG] host options conflict on Arch Linux


jrray:
On Arch Linux, our convention of using host option `"arch"` as the var to describe the CPU architecture conflicts with how we use the value of the `"distro"` var as the _key_ of another var, which is also `"arch"`, to describe the distro version.

We could rename `"arch"` to `"cpu"` and hope no Linux distribution takes that name, or come up with a way to namespace these vars to prevent this kind of collision. Or instead of dynamically naming this distro version var we use a fixed name like `"distro_version"`.

jrray:
As a bit of a bonus problem, there is no `VERSION_ID=` line in `/etc/os-release` (or any "rolling release" distro?) causing these two test failures:

```
thread 'cmd_ls::cmd_ls_test::test_ls_shows_remote_packages_with_host_default_filter_multiple_builds' panicked at crates/spk-cli/group2/src/./cmd_ls_test.rs:178:70:
called `Option::unwrap()` on a `None` value


failures:
    cmd_ls::cmd_ls_test::test_ls_shows_remote_packages_with_host_default_filter
    cmd_ls::cmd_ls_test::test_ls_shows_remote_packages_with_host_default_filter_multiple_builds
```

This is after I added renaming the detected distro to "archlinux" just so I could see if the tests would then all pass. Before the rename these tests were failing because the spk recipe was getting generated with two `"arch"` vars, but both with `"x86_64"` as the value:

```
 thread 'cmd_ls::cmd_ls_test::test_ls_shows_remote_packages_with_host_default_filter_multiple_builds' panicked at crates/spk-cli/group2/src/./cmd_ls_test.rs:171:16:                                                                                                                                                                                                   ▐ invalid spec: SerdeError { input: "{\"build\":{\"options\":[{\"var\":\"distro/arch\"},{\"var\":\"os/linux\"},{\"var\":\"arch/x86_64\"},{\"var\":\"arch/x86_64\"}]},\"pkg\":\"my-pkg/1.0.0/BGSHW3CN\"}", message: "build: build option was specified more than once: arch at line 1 column 10", line: Some(1), column: Some(9), contextualize: true, context_lines: 3, ▐ context_characters: 30 }
```

I thought that was from a name conflict but now I'm not sure, since there's no `VERSION_ID=` it means no (second) key called `"arch"` is being added to the host options. The name conflict is still a valid concern but that may not be the main problem here.

jrray:
[...]

PR #1219, opened 2025-05-22
Fix host options related ls tests that are failing on Arch Linux

jrray:
This is related to #1218 but doesn't resolve that issue, this makes the test suite pass when trying to run the tests on Arch Linux by "cheating" and setting the host option values explicitly for those tests. This PR is really just to get the code into a state where the test suite passes in my development environment.


crates/spk-schema/crates/foundation/src/option_map/mod.rs
```diff
@@ -101,10 +101,12 @@ impl HostOptions {
...
     /// the duration of the given future.
     ///
-    /// There is no guarantee that some other concurrent task doesn't also
-    /// change the options before the given future completes.
-    ///
     /// This method is intended to only be used by tests.
+    ///
+    /// There is no guarantee that some other concurrent task doesn't also
+    /// change the options before the given future completes. It is recommended
+    /// to use [`serial_test::serial`] and include the key "host_options" when
```
jrray, line 108: It seemed like a good idea to make a link here but it did require adding that crate to this module's dependencies.

rydrman (approved):

PR #1222, opened 2025-05-30
Bugfix: 0.44.x doesn't compile with sentry feature enabled

dcookspi:
This fixes a compile error in the Error::SpkSolverError matching in spk's cli error handling that trips when the`sentry` feature is enabled.


jrray (approved):
Thanks, shame we don't test this configuration in CI.

rydrman (approved):

PR #1226, opened 2025-06-07
Change `spk ls` to use  repo walkers

dcookspi:
This changes and refactors `spk ls` to use RepoWalkers. It reduced the number of cases inside `ls` while providing the same functionality. 

This is built on top of https://github.com/spkenv/spk/pull/1225

### DIFF
--- a/crates/spk-cli/group2/src/cmd_ls.rs
+++ b/crates/spk-cli/group2/src/cmd_ls.rs
@@ -205,10 +205,9 @@ impl<T: Output> Ls<T> {
         while let Some(item) = traversal.try_next().await? {
             if let RepoWalkerItem::Build(build) = item {
                 let package = build.spec.ident().name().to_string();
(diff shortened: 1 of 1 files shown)


crates/spk-cli/group2/src/cmd_ls.rs
```diff
@@ -314,197 +193,186 @@ impl<T: Output> CommandArgs for Ls<T> {
...
-                        Err(err) => {
-                            self.output.warn(format!("Skipping {build}: {err}"));
-                            continue;
-                        }
-                    };
-
-                    if !spec.matches_all_filters(filter_by) {
-                        continue;
-                    }
+                set.insert(package.clone());
```
jrray, line 211: This is traditionally done in a single step by inspecting the return value of `insert`.

jrray (approved):

PR #1228, opened 2025-06-07
Change `spk du` to use repo walkers

dcookspi:
This changes and refactors `spk du` to use RepoWalkers. The `du` command uses its own streams on top of the `RepoWalker`. This change leaves them inside the `du` command and not publically accessible. 
A future, pending, PR will refactor those further and move them into `spk-storage` for other commands to use.

This is built on top of #1225


crates/spk-cli/cmd-du/src/cmd_du.rs
```diff
@@ -289,258 +454,70 @@ impl<T: Output> Du<T> {
...
-                Err(e) => {
-                    self.output.warn(format_args!("{e}"));
-                    yield None
-                }
-            }
-        })
-    }
-
-    fn walk_components<'a>(
+    // TODO: move to the spk library so other things can use it?
```
dcookspi, line 457: This happens in another PR that is pending.

crates/spk-cli/cmd-du/src/cmd_du.rs
```diff
@@ -289,258 +454,70 @@ impl<T: Output> Du<T> {
...
+                .with_report_on_files(true)
+                .with_file_path(file_path)
+                .with_continue_on_error(true)
+                .build();
+
+            let mut traversal = repo_walker.walk();
+
+            let mut current_repo_name = Arc::new(String::from(""));
+            let mut pkg_name = Arc::new(PkgNameBuf::try_from("place-holder-package-name").unwrap());
+            let mut version = Arc::from(Version::new(0, 0, 0));
```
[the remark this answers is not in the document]
jrray: Using an Option here would let you initialize the variable as None. It would also make it possible to detect if the value is used before being assigned a real value, which could indicate a logic error. Or you can generate a default value on demand.

But don't change this for the sake of optimization because it won't make a meaningful difference.

crates/spk-cli/cmd-du/src/cmd_du_test.rs
```diff
@@ -123,40 +121,6 @@ async fn test_du_trivially_works(#[case] solver: SolverImpl) {
     assert_eq!(expected_output.len(), 0);
 }
 
-#[tokio::test]
-async fn test_du_warnings_when_object_is_tree_or_blob() {
```
jrray, line 127: Why is this test being deleted?

dcookspi: The original du command code returned an error if it encountered an spfs blob object. The new du command code both sets ignore errors on the repo walker, and the walker skips blobs entirely and uses spfs entries for files from manifests fir data. The error this was testing for is no longer generated, so the test isn't needed anymore.

crates/spk-solve/crates/solution/src/solution.rs
```diff
@@ -589,9 +627,30 @@ impl Solution {
...
             if verbosity > 0 {
+                let mut disk_usage = if size < 1024 {
+                    format!("{} ", spk_storage::human_readable(size))
+                } else {
+                    spk_storage::human_readable(size)
+                };
+                // Right aligning the size column by using a minimum
+                // width of 8 because the human readable format, after
+                // the adjustment above, is at least "ddd.d UU" = 8
+                let delta = 8 - disk_usage.len();
```
jrray, line 642: We discovered this can underflow.

Issue #1233, opened 2025-07-03
[BUG] spk build can skip capturing some files


jrray:
Now that we are using spk on a newer OS on a newer kernel we have observed a behavior change in overlayfs that causes a package build to not capture all the files that it used to on the older OS.

The problem stems from the built script using `mv` to move a directory from the source area into a new location under `/spfs`, and how we have optimized the changed file detection by looking at the overlayfs upperdir to quickly discover what was changed by the build.

An example case. Say the state of `/spfs` looks like this at the beginning of the build script:

- `/spfs/src/dir/a.txt`
- `/spfs/src/dir/b.txt`

This is the built script:

```
mkdir /spfs/dest
mv /spfs/src/dir /spfs/dest/here
```

After this executes, the overlayfs upperdir contains this:

- `src/dir` [mask]
- `dest/`
- `dest/here/`

With an older kernel, the two files would also appear here, but in the newer kernel they do not. On the newer kernel, the files `a.txt` and `b.txt` are not collected or appear in the built package.

One might come up with an idea to check for leaf/empty directories and then check the contents of the real overlayfs mount for those paths to find missing files. Unfortunately, it is possible for missing files to exist at any level of the tree. This can be demonstrated with a variation on the first example:

This is the built script:

```
mkdir /spfs/dest
mv /spfs/src/dir /spfs/dest/here
mkdir /spfs/dest/here/nested
```

After this executes, the overlayfs upperdir contains this:

- `src/dir` [mask]
- `dest/`
- `dest/here/`
- `dest/here/nested/`

The missing files are in the `here` directory which is neither empty or a leaf node.

None of the overlayfs options (index, metacopy) seem like it would make a difference in this behavior.

The old approach of using `spfs reset` and a full `spfs diff` was considerably slower than the technique we use today, but it can at least claim to be more correct in this case.

An outside the box idea is to make changes to how build environments are constructed so that the package `src` directory is actually a separate mount, so any `mv` opertions like this would cross a filesystem boundary and force those files to appear in the upperdir. It sounds like a "neat trick" but has some complex implementation requirements:

- spfs runtimes need to have knowledge of more than just one "stack" of layers
- each of these separate stacks need its own upperdir/workdir
- more than one mount to create / tear down
- src package contents have filenames like `spk/pkg/aaa/1.2.3/src` so the Manifest needs to strip the `spk/pkg/aaa/1.2.3` prefix (maybe we already have something like this)

[...]

PR #1238, opened 2025-07-17
Add redirect_dir properties for overlayfs

jrray:
To address #1233 start setting the redirect_dir setting when mounting an
overlayfs filesystem. Default to "follow" meaning the redirects won't be
created (but if any were to exist already for some reason they will be
honored). Redirects need to not be created to avoid the situation in the
issue where files are not collected because they are invisible in the
upperdir.

Through experimentation and noticing fine print in the documentation, it
is necessary to disable the other overlayfs options when using "follow"
mode.


jrray:
@rydrman can you take a look please?

rydrman (approved):

PR #1247, opened 2025-08-29
Resolvo fail in spk build for packages with embeds

jrray:
Consider this recipe:

```yaml
pkg: this/1.0.0
api: v0/package
build:
  script:
    - true
install:
  embedded:
    - pkg: that/1.0.0
```

Using `spk make-source` on this produces a package with an embedded package (abridged):

```yaml
api: v0/package
pkg: this/1.0.0/src
install:
  embedded:
  - api: v0/package
    pkg: that/1.0.0/embedded
  components:
  - name: src
```

Note it claims to embed the "embedded" build, whatever that means. The inclusion of an embedded package is assumed to be unintentional.

The resolvo solver, when it picks the src build, would end up creating a transitive dependency on `local/that/==1.0.0/embedded[this:run/1.0.0/src]` (the run component) and this will generally not exist yet, since the `this` package is in the process of being built.

If instead resolvo would try to make the dependency on `local/that/==1.0.0/embedded[this:src/1.0.0/src]` (the src component), it would create an unresolvable cycle.

It could be argued that the dependency cycle should just cancel out and be skipped, and that's what happens here since the code now just ignores embedded packages when resolving a src package.

A test was added that reproduced the problem before the fix.

### DIFF
--- a/crates/spk-solve/src/solvers/resolvo/spk_provider.rs
+++ b/crates/spk-solve/src/solvers/resolvo/spk_provider.rs
@@ -1424,93 +1424,104 @@ impl DependencyProvider for SpkProvider {
                     }
                 }
                 // Also add dependencies on any packages embedded in this
-                // component.
-                for embedded in package.embedded().iter() {
-                    // If this embedded package is configured to exist in
-                    // specific components, then skip it if this solvable's
-                    // component is not one of those.
-                    let components_where_this_embedded_package_exists = package
-                        .components()
-                        .iter()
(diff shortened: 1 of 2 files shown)


rydrman:
From the meeting today:
- as part of this, we should be stripping generated source package specs of all (at least root level) embedded packages since they are nonsensical unless the spec is being hand-written for some reason (though we have no current use case for that)

rydrman (approved):

PR #1251, opened 2025-09-05
Use matrix builds to check various feature combinations

jrray:
Use github matrix builds for more coverage in CI for different feature configurations, in particular the one used at SPI.

The problems found by using cargo-feature-combinations locally were partially due to crates not propagating the "statsd" feature to dependencies. Other problems come from code that assumes default features will be enabled but don't compile successfully when they aren't. These have been identified in the code with comments.

[This crate](https://crates.io/crates/featurecomb) could be used to enforce features are enabled, and would be useful for the vfs/fuse crates that don't work unless one of the fuse abi features is enabled.

Everything is passing with `cargo fc --exclude-package spfs-vfs --exclude-package spfs-fuse --exclude-package spfs-cli-fuse check`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -140,3 +140,68 @@ jobs:
         run: |
           # Run tests that need root
           crates/spfs/tests/integration/run_privileged_tests.sh
+
+  matrix-build:
+    name: build features ${{ matrix.features }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    container:
+      image: rust:slim-bookworm
+      options: --security-opt seccomp=unconfined --privileged
+    env:
+      # Enable sccache for rust
+      SCCACHE_GHA_ENABLED: "true"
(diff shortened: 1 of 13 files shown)


.github/workflows/rust.yml
```diff
@@ -140,3 +140,70 @@ jobs:
...
+      RUSTC_WRAPPER: "sccache"
+    strategy:
+      fail-fast: false
+      matrix:
+        features:
+          # SPI's standand features
+          - migration-to-components,sentry,spfs/protobuf-src,spfs-vfs/protobuf-src,statsd,fuse-backend-rhel-7-6,legacy-spk-version-tags
+          # This feature is known to have problems compiling if other features
+          # are not enabled.
+          - statsd
```
jrray, line 163: @rydrman is there a particular feature combination you would like to see added here or does the preexisting CI jobs cover your use case already?

rydrman (approved):

rydrman (approved):

PR #1259, opened 2025-09-11
Restore coverage workflow but using grcov

dcookspi:
This restores the coverage CI workflow and changes it to use grcov for the coverage reporting (via `make coverage`). This also add a `coverage` rule to the Makefile.


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,86 @@
...
+        run: |
+          apt-get update
+          apt-get install --yes build-essential libcap2-bin sudo cmake tcsh rsync protobuf-compiler fuse libfuse-dev curl unzip pkg-config libssl-dev git
+          # spfs-fuse requires this option enabled
+          echo user_allow_other >> /etc/fuse.conf
+          FB_REL=https://github.com/google/flatbuffers/releases/
+          curl --proto '=https' --tlsv1.2 -sSfL ${FB_REL}/download/v23.5.26/Linux.flatc.binary.g++-10.zip | funzip > /usr/bin/flatc
+          chmod +x /usr/bin/flatc
+          rustup install nightly
+          rustup component add rustfmt --toolchain nightly
```
jrray, line 49: This workflow doesn't run any linting/fmt tasks so installing these can probably be skipped.

jrray (approved):
Might be able to save a little bit more, but otherwise LGTM.

PR #1268, opened 2025-09-24
Test solver behavior when using options

jrray:
Is it expected that the solver will make an effort to resolve packages that fit the options, even though it is not required to. Create a test where multiple builds exist with different values for a var and add an option to the solver for a specific value. A build exists with that value so the solver should pick it.


jrray:
I labeled this with `bug` because the step solver doesn't pass the test. First things first, do we believe this test is valid and should pass?

PR #1274, opened 2025-09-26
Remove legacy-spk-version-tags feature

jrray:
This follows up #1196 and removes the legacy tags support originally added in #972.

SPI required this for a time while we were in the process of fixing all the non-normalized version tags in our repository, and updating to a build of spk that only wrote normalized tags.

Any other contemporary users of spk should not have any non-normalized tags to begin with.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -157,7 +157,7 @@ jobs:
       matrix:
         features:
           # SPI's standand features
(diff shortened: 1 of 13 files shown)


rydrman (approved):

rydrman (approved):
This is a tough diff, nothing obvious to me, but I am also trusting the existing tests on this one

PR #1278, opened 2025-10-03
Prevent FallbackRepository from download storms

jrray:
Use a moka Cache to funnel multiple requests for the same missing payload into a single task to sync the payload locally. Any failure to sync the payload will be cached for 5 minutes, after which another request for that digest will attempt to sync it again.

This was inspired by observing that syncing a single spfs tag would report having "repaired" several different digests multiple times, often over a dozen times. It suggests that Syncer may unexpectedly trying to process the same digest multiple times. Each of these digests were for large payloads (500MB+). The sync was happening over a high-latency link. Without this change, it can end up performing excessive duplicate copies over a limited bandwidth link.

### DIFF
--- a/crates/spfs/src/storage/fallback/repository.rs
+++ b/crates/spfs/src/storage/fallback/repository.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
 use futures::Stream;
+use moka::future::Cache;
 use relative_path::RelativePath;
 
 use crate::config::{ToAddress, default_fallback_repo_include_secondary_tags};
@@ -85,6 +86,7 @@ pub struct FallbackProxy {
     primary: Arc<OpenFsRepository>,
     secondary: Vec<crate::storage::RepositoryHandle>,
     include_secondary_tags: bool,
+    open_payload_cache: Cache<encoding::Digest, ()>,
 }
 
 impl FallbackProxy {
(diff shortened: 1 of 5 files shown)


crates/spfs/src/error.rs
```diff
@@ -384,3 +387,39 @@ impl OsError for std::io::Error {
...
 pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Diagnostic, Debug, Clone, Error)]
+#[diagnostic(
+    url(
+        "https://spkenv.dev/error_codes#{}",
+        self.code().unwrap_or_else(|| Box::new("spfs::generic"))
+    )
+)]
+pub enum SyncError {
```
jrray, line 398: I did all this Error rework and then realized I didn't need it. But I think this change is for the better anyway.

crates/spfs/src/storage/fallback/repository.rs
```diff
@@ -297,63 +305,66 @@ impl PayloadStorage for FallbackProxy {
...
-            return Err(missing_payload_error);
-        }
+                if let Some(err) = repair_failure {
+                    return Err(err);
+                }
+                // Probably can only get here if there are no secondary repos.
+                Err(SyncError::PayloadReadError(digest, Error::String("no repositories could successfully read the payload".into()).into()))
+            })
+            .await
+            .map_err(|err| Error::String(format!("failed to open payload: {err}")))?;
```
jrray, line 363: I would rather return a more specific error here, but this function has to return spfs::Error and in error type in the cache is now SyncError. Most of the SyncError variants have a nested Arc<spfs::Error> that in theory could be returned, but spfs::Error is not Clone.

I worked on making a new error type that is Clone but oops, this function would have to change to return that error type. Or it could change to return yet another new Error type that is Clone and more specific to the Repository domain instead of the catch-all spfs::Error type.

PR #1282, opened 2025-10-21
Demonstrate spfs clean can corrupt repository

jrray:
Clean can end up violating the invariant that all child objects of a parent must exist. This test demonstrates this with a manifest and a layer but the problem applies to other relationships as well.


jrray:
Closing now that #1288 can pass this test.

PR #1286, opened 2025-10-23
Adds test specs access from a build spec

dcookspi:
This adds  `get_all_tests()` method to the Package, Spec etc. APIs to allow a build's test specs to be accessed from a build spec (a v0::Spec). Build spec's have the data aready, but accessing it is a bit convoluted without these changes.

### DIFF
--- a/crates/spk-schema/src/package.rs
+++ b/crates/spk-schema/src/package.rs
@@ -14,6 +14,7 @@ use spk_schema_foundation::version::VERSION_SEP;
 use super::RequirementsList;
 use crate::foundation::ident_component::Component;
 use crate::foundation::option_map::OptionMap;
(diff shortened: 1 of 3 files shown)


crates/spk-schema/src/v0/spec.rs
```diff
@@ -343,6 +344,10 @@ impl Package for Spec<BuildIdent> {
         Cow::Borrowed(&self.install.requirements)
     }
 
+    fn get_all_tests(&self) -> Vec<SpecTest> {
+        self.tests.clone().into_iter().map(SpecTest::V0).collect()
```
jrray: `TestSpec` to `SpecTest`? 🙃

PR #1289, opened 2025-11-11
Ambiguous digest problems with existing blob objects

jrray:
After testing with the blob removal against a repo that contains blob objects it became apparent that it was impossible to do things like `spfs cat` the payload digest (both using a full or a partial digest) because it would fail saying the digest is ambiguous.

    $ spfs info PWDF5FM3ERTJDDEYMOX4VFBNB64J27E2YDEZXL6DOSKQJXWZO4YA====
    Error:   × Ambiguous reference [too short]: PWDF5FM3ERTJDDEYMOX4VFBNB64J27E2YDEZXL6DOSKQJXWZO4YA

My first thought was that this situation is _not_ ambiguous and implemented a way to make it prefer the payload when a blob object exists. However, on second thought I realized that it is possible for a non-blob object to share a digest with an unrelated payload (granted this is highly improbable). I decided to make it so the caller can hint at what type of item is expected when resolving a partial digest.

I think it is for the greater good to prefer payloads as the default behavior when the type is unknown, but now there is still a way to resolve ambiguities for this other extreme edge case. It didn't seem appropriate to try to read the object to check if it is a blob and disambiguate that way, but that would be an option.

### DIFF
--- a/crates/spfs/src/graph/database.rs
+++ b/crates/spfs/src/graph/database.rs
@@ -404,12 +404,34 @@ impl<T: Database> Database for &T {
 #[async_trait::async_trait]
 pub trait DatabaseExt: Send + Sync {
     /// Write an object to the database, for later retrieval.
+    ///
+    /// It is not permitted to write blob objects.
     async fn write_object<T: ObjectProto>(&self, obj: &FlatObject<T>) -> Result<()>;
+
+    /// Write an object to the database, for later retrieval.
+    ///
+    /// # Safety
+    ///
+    /// This function does not check the type of the object being written. It
+    /// is expected that the caller will not write a blob object except in
+    /// specific cases, such as in test code.
+    async unsafe fn write_object_unchecked<T: ObjectProto>(
+        &self,
+        obj: &FlatObject<T>,
+    ) -> Result<()>;
 }
 
 #[async_trait::async_trait]
 impl<T: DatabaseExt + Send + Sync> DatabaseExt for &T {
     async fn write_object<O: ObjectProto>(&self, obj: &FlatObject<O>) -> Result<()> {
         DatabaseExt::write_object(&**self, obj).await
+    }
+
+    async unsafe fn write_object_unchecked<O: ObjectProto>(
+        &self,
+        obj: &FlatObject<O>,
+    ) -> Result<()> {
+        // Safety: transitive unsafe call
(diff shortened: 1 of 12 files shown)


rydrman:
Am I wrong in thinking that we had some kind of salting process for objects so that we wouldn't need to practically worry about the kind of collision that you're addressing with the hinting?

jrray:
As an exercise I modified the flatbuffer object digest calculation to save all the bytes that pass through the hasher.

The test called `storage::repository::database_test::test_object_existence` writes a layer object with digest `NCDBKSOXIJFJDENWSKY5OPKAWMYUGAG32KFMHEHOWGTRLZWDWGTA====`.

With my modifications I get a file called `digest-NCDBKSOXIJFJDENWSKY5OPKAWMYUGAG32KFMHEHOWGTRLZWDWGTA====.bin`. The contents of this file are the bytes that were hashed to calculate the object digest. If I put this file into an spfs layer then I get a payload with the same digest as the object:
```
$ spfs info D3QJEFHQRMM24WQXOSXA3F6WXJHPFZNU75G57UJLGUZU4IH3PQHQ====
D3QJEFHQRMM24WQXOSXA3F6WXJHPFZNU75G57UJLGUZU4IH3PQHQ====:
manifest:
 100644 file NCDBKSOXIJFJDENWSKY5OPKAWMYUGAG32KFMHEHOWGTRLZWDWGTA==== /spfs/digest-NCDBKSOXIJFJDENWSKY5OPKAWMYUGAG32KFMHEHOWGTRLZWDWGTA====.bin
```

So it is a contrived example but shows how it is possible to have a non-blob object and payload share the same digest. A payload contains arbitrary bytes so all possible digest values are possible.

I was thinking about using this technique to generate content for tests for coverage of when there is a digest collision in a repo. But, we don't require payload contents actually hash correctly so test content could be cheesed by manually writing a payload with a desired digest without proper file contents.

But I don't think this is something we need to worry about. A change that could help is making a decision that an `EnvSpecItem` can't be a payload digest. Trouble is that that type is used is contexts where a payload is acceptable.

- `spfs push <payload digest>` ✅ 
- `spfs shell <payload digest>` ❌ 

There could be a different type that is a superset of `EnvSpecItem` that does allow payload digests but there would be limited places that accept this type.

PR #1293, opened 2025-11-22
Remove validate_options

jrray:
This has been dead code since #530, which deleted the last use of it, which was [here](https://github.com/spkenv/spk/blob/eff0825a6a0f99a24696a903687c230b4908acea/crates/spk-schema/src/v0/spec.rs#L323).


jrray:
I don't know if it _should_ be dead code, but as you can see all our tests pass without it. We can bring it back if someone can propose a test that reveals it is needed.

rydrman (approved):

PR #1295, opened 2025-11-25
Replace Spec with RecipeSpec and PackageSpec

jrray:
This is the first step towards having separate types for the different variations of specs that are valid in specific contexts. These changes will be done in a series of PRs to limit the diff size of each change.

In this PR the focus is on embedded packages. In a recipe, an embedded package spec in a recipe spec is expected to be given a `VersionIdent`, as in:

```yaml
install:
  embedded:
    - pkg: example/1.0.0
```

It is now illegal to specify a build id for an embedded package in a recipe.

When the package is built, the embedded package in a package spec is a `BuildIdent` to reflect the idea that the embedded package is "built", such as:

```yaml
install:
  embedded:
    - pkg: example/1.0.0/embedded
```

Test content has been updated appropriately since this was not enforced previously.

What is allowed in the `build:` and `install:` sections of an embedded package is different from a regular package, so `EmbeddedRecipeSpec` and `EmbeddedPackageSpec` model those differences. These contain `EmbeddedBuildSpec` and `EmbeddedInstallSpec`.

An `EmbeddedInstallSpec` does not contain an `embedded` field, making it impossible to nest embedded packages.

`EmbeddedPackagesList` gained a generic parameter for the type of spec it contains so it can be reused for the two flavors of embedded specs.

`Package::components()` was moved to its own trait to make is possible to not have `Package` implemented for embedded packages. Note that when embedded packages are turned into stubs, the stubs are normal packages that implement `Package`.

All the new types were created in the `v0` module since they are part of the `v0/package` API. The types like `BuildSpec` and `InstallSpec` should arguably have been defined in `v0` as well.

These new types were created by copying and modifying the existing types (and tests) and I haven't (at this point) done the work to remove duplicate or unused code.

### DIFF
--- a/crates/spk-schema/src/v0/embedded_recipe_spec.rs
+++ b/crates/spk-schema/src/v0/embedded_recipe_spec.rs
@@ -2,19 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/spkenv/spk
 
-use std::borrow::Cow;
-use std::str::FromStr;
-
 use serde::{Deserialize, Serialize};
 use spk_schema_foundation::IsDefault;
 use spk_schema_foundation::ident::{AsVersionIdent, VersionIdent};
-use spk_schema_foundation::version::{IncompatibleReason, VarOptionProblem};
 
 use super::TestSpec;
 use crate::foundation::name::PkgName;
 use crate::foundation::spec_ops::prelude::*;
-use crate::foundation::version::{Compat, Compatibility, Version};
-use crate::ident::{Satisfy, VarRequest, is_false};
+use crate::foundation::version::{Compat, Version};
+use crate::ident::is_false;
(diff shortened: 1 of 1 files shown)


crates/spk-schema/src/v0/package_spec.rs
```diff
@@ -0,0 +1,668 @@
...
+    SourceSpec,
+    ValidationSpec,
+};
+
+#[cfg(test)]
+#[path = "./package_spec_test.rs"]
+mod package_spec_test;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, Ord, PartialOrd, Serialize)]
+pub struct PackageSpec {
```
rydrman: just noting that some of these "top-level" structs don't have any descriptions, might be a useful addition if you are feeling inclined

PR #1301, opened 2025-12-07
Hotfix packages with unpinned var opts on load

jrray:
Given a package with an embedded package that defines a var with a
default value, such as in:

```yaml
install:
  embedded:
    - pkg: python/3.10.8
      build:
        options:
          - var: abi/3.10.8
```

Until the fix in #1302, these vars were not getting pinned
when the package is built and the stub created:

```yaml
pkg: python/3.10.8/embedded[some-parent-pkg:run/3.10.8/P7SZECZD]
api: v0/package
build:
  options:
  - var: python.abi/cp310
```

This created a problem for solving correctly because
`Satisfy<VarRequest>` would incorrectly claim that this package is compatible
with _any_ value for `python.abi`.

In order to fix solves for the existing embedded stubs, these missing
pinned values are pinned as the package specs are read from storage, so
they end up with the intended content:

```yaml
pkg: python/3.10.8/embedded[some-parent-pkg:run/3.10.8/P7SZECZD]
api: v0/package
build:
  options:
  - var: python.abi/cp310
    static: cp310
```

Now `Satisfy<VarRequest>` will correctly reject this package as not
being compatible with, e.g., `python.abi/cp311`.

The best we can do is assume `python.abi` was not overridden to some
other value at the time the parent package was built and that the
default on the option is the appropriate value for the pin.

### DIFF
--- a/crates/spk-storage/src/storage/spfs.rs
+++ b/crates/spk-storage/src/storage/spfs.rs
@@ -26,7 +26,7 @@ use spk_schema::ident::{AsVersionIdent, VersionIdent};
 use spk_schema::ident_build::parsing::embedded_source_package;
 use spk_schema::ident_build::{EmbeddedSource, EmbeddedSourcePackage};
 use spk_schema::ident_ops::TagPath;
-use spk_schema::{AnyIdent, BuildIdent, FromYaml, Package, Recipe, Spec, SpecRecipe};
+use spk_schema::{AnyIdent, BuildIdent, FromYaml, Opt, Package, Recipe, Spec, SpecRecipe};
 use tokio::io::AsyncReadExt;
 
 use super::CachePolicy;
@@ -504,7 +504,6 @@ impl Storage for SpfsRepository {
         {
             return v.value().clone().into();
         }
-
         let r: Result<Arc<Spec>> = async move {
             let tag_path = Self::build_spec_tag(pkg);
             let tag_spec = spfs::tracking::TagSpec::parse(tag_path.as_str())?;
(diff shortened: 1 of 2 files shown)


jrray:
Some extra commentary... In the spirit of #1298 and friends, IMO it will help with bugs like this if/when an embedded package has a type that mandates that build options are pinned.

You could take things further since "build options" are arguably meaningless for something that can never be built: an embedded package.

You can also argue that a package, embedded or not, should not have a "build" section at all anymore, and anything from the "build" section of a recipe that is needed at runtime should be populated into the "install" section of the package at build time.

Similarly, an embedded package could arguably not have an "install" section and any of its runtime requirements should live in the runtime requirements of the component in the parent package that embeds this package.

Changes like these would go a long way towards simplifying how package dependencies are calculated and reduce the opportunity for bugs to sneak in.

For example, `Package::runtime_requirements()` returns a reference to the `install.requirements` field, The name might suggest that's all you need to figure out the runtime requirements of a package, however you still need to look at [some of] the `build.options`, components, embedded packages...

Handling components could be made easier if we baked out package specs for each component and treated them like separate packages. The `InstallSpec` of a built package would not have a `components` field anymore. It could also not have an `embedded` section anymore, just pkg requests for the stub package. The end goal being that `install.requirements` does become all you need to know the runtime requirements.

crates/spk-storage/src/storage/spfs.rs
```diff
@@ -517,6 +516,29 @@ impl Storage for SpfsRepository {
                 .await
                 .map_err(|err| Error::FileReadError(filename, err))?;
             Spec::from_yaml(&yaml)
+                .map(|spec| match spec {
```
jrray: I can understand that this might be undesirable, but it is needed at least in the short term to fix bad solves. With the other fix in this PR in place, we can work on republishing packages to fix their stubs (or write some fixer script), and then I would be good with reverting this.

rydrman (approved):

PR #1303, opened 2025-12-12
Fix VarOpt Hash and Ord implementations

jrray:
The compat field was not being considered. Use destructuring to prevent this kind of error in the future.

In Ord, `self.description` was being compared to `other.value`. It didn't matter because the result was thrown away. However, since `description` is included in the Hash, it should also be included in the Ord implementation to maintain consistency. No justification for it being skipped was given in #954.

### DIFF
--- a/crates/spk-schema/src/option.rs
+++ b/crates/spk-schema/src/option.rs
@@ -335,6 +335,7 @@ impl<'de> Deserialize<'de> for Opt {
 pub struct VarOpt {
     pub var: OptNameBuf,
     pub default: String,
+    // This field does not implement `Ord` so this struct can't derive it.
     pub choices: IndexSet<String>,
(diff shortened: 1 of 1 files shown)


crates/spk-schema/src/option.rs
```diff
@@ -346,39 +346,67 @@ pub struct VarOpt {
...
-        self.description.hash(state);
-        self.value.hash(state)
+        inheritance.hash(state);
+        description.hash(state);
+        compat.hash(state);
+        value.hash(state)
     }
 }
 
 impl Ord for VarOpt {
```
jrray: I'm not seeing a reason why this can't be derived. I don't see why it wasn't derived when originally added in #377, it may have come down to inexperience with Rust at the time.

PR #1314, opened 2026-01-22
Support roundtrip of embedded stub build tags

jrray:
A side effect of #1306 is that it broke `spk ls` and maybe other things in the event that an embedded stub was created in the past with a non-normalized version number.

The buildid of an embedded stub is a base32 encoded string that points at the parent of the embedded package, like in:

    embedded[my-pkg:run/2024.2.2.29.0/K76MI2A4]

This example contains a non-normalized version number, and when parsed it would be normalized to:

    embedded[my-pkg:run/2024.2.2.29/K76MI2A4]

Unfortunately that means if this buildid is then used to recreate the expected tag name of the embedded stub it will not match the original tag name (aka "metadata" name) as read from storage.

This is resolved by capturing the original string as an additional field in the `EmbeddedSourcePackage` when parsing it that can be used to get back the original tag name when requested.

Yes, this works around what is now considered "illegal" content in the repo but it is nice to provide continuity when there is no easy way to find and fix this problem in a repo.

There remains another related problem that operations on the parent package will fail to find the existing embedded stub that contains a non-normalized version. A certain amount of manual intervention will be needed to fix these issues in the repo.

A different approach with similar outcomes would have been to capture the non-normalized version string in a new field in `IdentPartsBuf`, but there isn't a known use for that outside this particular case.


jrray:
@rydrman this is the fix needed for the problem I mentioned today, along with newly discovered #1313. These both are needed to avoid errors when trying to read existing content in our repo.

rydrman (commented):
Seems reasonable to me

rydrman (approved):

PR #1319, opened 2026-02-02
Add note about auto_host_vars to build page

rydrman:


jrray (approved):
Does this have to be a single long line for the note styling to work?

PR #1337, opened 2026-03-19
Indexes 2: Add `new_unchecked()` constructors to spk schema objects

dcookspi:
_Note: for info on benefits of indexing for spk solves see https://github.com/spkenv/spk/pull/1340 (5 of 5). Maybe start there and work back down to this PR if you prefer to review PRs top down._

This adds `new_unchecked()` constructor methods to various spk schema objects. These allow direct creation of those objects from other existing object pieces, e.g. from other pieces of those objects in index data. This is one of the changes that supports adding indexes and index based packages to Spk repositories. It allows indexes to avoid reparsing data from text for some objects.

This is 2 of 5 chained PRs for adding indexes to spk solves:
1. https://github.com/spkenv/spk/pull/1336
2. this PR
3. https://github.com/spkenv/spk/pull/1338
4. https://github.com/spkenv/spk/pull/1339
5. https://github.com/spkenv/spk/pull/1340
6. #1344
7. #1354
8. #1355

PR #1338, opened 2026-03-19
Indexes 3: Adds flatbuffers schema and SolverPackageSpec for indexes to spk

dcookspi:
_Note: for info on benefits of indexing for spk solves see https://github.com/spkenv/spk/pull/1340 (5 of 5). Maybe start there and work back down to this PR if you prefer to review PRs top down._

This adds a new `spk-proto` crate to spk, and a new v0 package implementation: `SolverPackageSpec`. `spk-proto` contains a flatbuffers schema for index package, version, build, options, and global variables data.  The flatbuffers schema is designed for solver use. It only stores data needed for solving with the packages. It is not a complete replacement for all package operations. It does not have enough data to build or test packages.

`SolverPackageSpec` is a package implementation backed by a flatbuffer. Those packages will be produced from an index by changes added in a subsequent PR (https://github.com/spkenv/spk/pull/1339, 4 of 5).  The functions for converting spk schema objects into and out of flatbuffer objects are contained in `fb_converter.rs`. 

This is one of the changes that supports adding indexes and index based packages to Spk repositories. 

This is 3 of 5 chained PRs for adding indexes to spk solves:
1. https://github.com/spkenv/spk/pull/1336
2. https://github.com/spkenv/spk/pull/1337
3. this PR
4. https://github.com/spkenv/spk/pull/1339
5. https://github.com/spkenv/spk/pull/1340
6. https://github.com/spkenv/spk/pull/1344
7. #1354
8. #1355


crates/spk-proto/schema/spk.fbs
```diff
@@ -0,0 +1,327 @@
...
+table TagSetItem {
+    name: string (required);
+    number: uint32;
+}
+
+
+// The pieces of a pkg/version/build Ident, but not combined as a
+// BuildIdent. This is used for embedded packages.
+table IdentPartsBuf {
+     repository_name: string;
```
jrray, line 47: I can't think of a reason why this field would ever [need to be] populated in the index. But if there's zero cost to having it here...

crates/spk-proto/schema/spk.fbs
```diff
@@ -0,0 +1,327 @@
...
+   // install: EmbeddedInstallSpec - mostly ignored, flattened into only this field,
+   component_specs: [SolverComponentSpec];
+   // requirements - field ignored in favour of install requirements with options,
+   //                which are stored in the requirements field below
+   requirements: [RequirementWithOptions];
+}
+
+// A Package/Version/Build containing what is needed when solving
+table BuildIndex {
+   // Older packages may have been published with non-normalized
```
jrray: This isn't really about "older packages" but that this preserves the version number as written in the recipe (to please users) versus the normalized version number which is used for comparisons.

crates/spk-proto/schema/spk.fbs
```diff
@@ -0,0 +1,327 @@
...
+    epsilon: Epsilon;
+    pre: [TagSetItem];
+    post: [TagSetItem];
+}
+
+// A Package/Version
+table VersionIndex {
+    // The version number only
+    version: Version (required);
+    builds: [BuildIndex];
```
jrray, line 303: Are builds sorted?

crates/spk-proto/schema/spk.fbs
```diff
@@ -0,0 +1,327 @@
...
+// For global var values
+table GlobalVar {
+    // An OptNameBuf
+    name: string (required);
+    // The set of known values for this global variable
+    values: [string];
+}
+
+// The index for a repository
+table RepositoryIndex {
```
jrray, line 323: Let's at least add some field that represents a high level version number that governs the practices outlined in comments but aren't represented in the types. If we change validation rules, expectations, etc., then we would bump this number up.

crates/spk-proto/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[dependencies]
+data-encoding = { workspace = true }
+flatbuffers = { workspace = true }
+futures = { workspace = true }
+miette = { workspace = true }
+serde = { workspace = true, optional = true }
+thiserror = { workspace = true }
+
+[build-dependencies]
+flatc-rust = "0.2"
```
jrray: Please move this version up to workspace and update spfs-proto to use the workspace version too.

crates/spk-schema/src/v0/solver_package_spec.rs
```diff
@@ -0,0 +1,547 @@
...
+    fb_embedded_package_specs_to_embedded_package_specs,
+    fb_opt_to_opt,
+    fb_opts_to_opts,
+};
+
+// TODO: should this be IndexedPackageSpec or something else? It comes
+// from an index and requires one to function, but it is used/usable
+// only by the solver.
+#[derive(Debug, Serialize)]
+pub struct SolverPackageSpec {
```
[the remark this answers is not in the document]
jrray: First I'm not sure we even want to use "spec" in the name here, since to me that word makes me think more about content a human would write and/or yaml-formatted content, based on all the existing things we refer to as a "spec".

I'll propose `IndexedPackage` as a name for this.

I don't think it is true that it is only used by the solver. We already have plans to use the index for other kinds of reporting.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+    #[serde(skip)]
+    cached_compat: ArcSwapOption<Compat>,
+    #[serde(skip)]
+    cached_embedded: ArcSwap<EmbeddedPackagesList<EmbeddedPackageSpec>>,
+    #[serde(skip)]
+    cached_component_specs: ArcSwap<ComponentSpecList<ComponentSpec>>,
+    #[serde(skip)]
+    cached_requirements: ArcSwap<RequirementsList<RequestWithOptions>>,
+    #[serde(skip)]
+    cached_build_opts: ArcSwap<Vec<Opt>>,
```
jrray, line 81: I'd like to discuss if `moka` is a better fit for these caches.

crates/spk-schema/src/v0/solver_package_spec.rs
```diff
@@ -0,0 +1,547 @@
...
+            cached_build_opts: ArcSwap::new(self.cached_build_opts.load_full()),
+        }
+    }
+}
+
+impl std::hash::Hash for SolverPackageSpec {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.build_ident.hash(state);
+        // TODO: skip this? It could be quite large?
+        self.buf.hash(state);
```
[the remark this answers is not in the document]
jrray: It would presumably be way cheaper to only hash the offset. That should be enough to cover uniqueness, I'm assuming no two packages in the index are going to have the same offset.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+        self.offset.hash(state);
+        // The cached fields are skipped
+    }
+}
+
+impl std::cmp::PartialEq for SolverPackageSpec {
+    fn eq(&self, other: &Self) -> bool {
+        // This deliberately ignores the buf field, offset field, and
+        // all the caches.
+        self.build_ident == other.build_ident && self.offset == other.offset
```
jrray: It may be a micro-optimization to test offset first here, if testing anything other than offset is even necessary. The logic for PartialEq and Hash ought to agree.

crates/spk-schema/src/v0/solver_package_spec.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+impl Deprecate for SolverPackageSpec {
+    fn is_deprecated(&self) -> bool {
+        self.build_index().is_deprecated()
+    }
+}
+
+impl DeprecateMut for SolverPackageSpec {
```
jrray: sus if the solver requires this trait, but that's a battle for another time.

crates/spk-schema/src/v0/solver_package_spec.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+impl Named for SolverPackageSpec {
+    fn name(&self) -> &PkgName {
+        self.build_ident.name()
+    }
+}
+
+impl RuntimeEnvironment for SolverPackageSpec {
```
jrray: Add this to the list of traits to figure out why the solver would need it implemented.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+impl Versioned for SolverPackageSpec {
+    fn compat(&self) -> Cow<'_, Compat> {
+        if self.cached_compat.load().is_none() {
+            // The compat hasn't been read and decoded yet
+            let compat = fb_compat_to_compat(self.build_index().compat());
+            self.cached_compat.store(Some(Arc::new(compat)))
+        }
```
jrray, line 338: To go with my comment about `moka` above, it would in theory be possible to set up the cache for it to compute these values on first access, so the hope would be it would simplify or even eliminate a lot of the code currently required at every place the cache is accessed.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,547 @@
...
+    fn ident(&self) -> &BuildIdent {
+        &self.build_ident
+    }
+
+    fn metadata(&self) -> &crate::metadata::Meta {
+        let err =
+            Error::SpkIndexPackageDoesNotImplement("Package".to_string(), "metadata".to_string());
+        // TODO: should this change the return value, update all the
+        // caller's handling, and return an error for this implementation?
+        unreachable!("{err}");
```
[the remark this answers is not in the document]
jrray: It's more work of course but I would like to refactor traits such that `metadata` isn't a required method on `Package` and then also the solver doesn't require an implementation of whatever [new] trait `metadata` moves to. Therefore you know for a fact this is unreachable; the code doesn't have to exist.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    fn matches_all_filters(&self, filter_by: &Option<Vec<OptFilter>>) -> bool {
+        // A copy of the code in V0 Spec's matchers_all_filters
```
jrray: If this is a literal copy, unchanged, then it could live in an extension trait that both these types share.

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+                if !compat.is_ok() {
+                    return false;
+                }
+            }
+        }
+        // All the filters match, or there were no filters
+        true
+    }
+
+    fn sources(&self) -> &Vec<SourceSpec> {
```
jrray, line 414: Similar to the idea above about moving `metadata` out of `Package`, I'm thinking `Package` shouldn't have a `sources` either -- wouldn't that be a property for `Recipe` to have? In the future with #1187 we won't have this property in a "package".

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -0,0 +1,545 @@
...
+            let reqs = fb_requirements_to_requirements(self.build_index().runtime_requirements());
+            let requirements = RequirementsList::<RequestWithOptions>::new_checked(reqs);
+
+            self.cached_requirements.store(Arc::new(requirements));
+        }
+
+        Cow::Owned((**self.cached_requirements.load()).clone())
+    }
+
+    fn get_all_tests(&self) -> Vec<SpecTest> {
```
jrray, line 475: Same idea here of something that shouldn't be in `Package`.

It does leave me wondering if we want to have a separate `SolverPackage` trait with a reduced set of requirements that the solver expects. We either need that or some combination of sub-dividing `Package` to cater to the different needs of different subsystems.

crates/spk-schema/src/fb_converter.rs
```diff
@@ -0,0 +1,1358 @@
...
+        }
+    };
+}
+
+// TODO: all these functions are here to keep the flatbuffer
+// conversions together while deciding on whether to replace the
+// backing of current rust objects with flatbuffer ones, like
+// SolverPackageSpec does. Or whether to split these functions up into
+// a couple of traits (e.g. IntoFlatbuffer, FromFlatbuffer) across
+// each of the existing rust types.
```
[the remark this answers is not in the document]
jrray: I like the idea of leaving them all here in one place as a starting point.

crates/spk-schema/src/fb_converter.rs
```diff
@@ -0,0 +1,1364 @@
...
+
+#[inline]
+pub fn fb_prerelease_policy_to_prerelease_policy(
+    fb_prerelease_policy: spk_proto::PreReleasePolicy,
+) -> Option<PreReleasePolicy> {
+    match fb_prerelease_policy {
+        spk_proto::PreReleasePolicy::None => None,
+        spk_proto::PreReleasePolicy::ExcludeAll => Some(PreReleasePolicy::ExcludeAll),
+        spk_proto::PreReleasePolicy::IncludeAll => Some(PreReleasePolicy::IncludeAll),
+        _ => None,
```
jrray: are these enums generated with non-exhaustive?

jrray: I guess based on some code comments below that these are plain numbers.

It could be a good idea to add `debug_assert!()`s on each of these catch all cases to detect things that were missed after any schema changes.

crates/spk-schema/src/fb_converter.rs
```diff
@@ -0,0 +1,1364 @@
...
+                    .map(|v| fb_version_to_version(v));
+                let ident = OptVersionIdent::new(name, version);
+
+                let components =
+                    fb_component_names_to_component_names_set(&component_embedded_pkg.components());
+
+                ComponentEmbeddedPackage::new_unchecked(ident, components)
+            })
+            .collect()
+        // TODO: This does not set fabricated, not sure if that is a problem or not?
```
jrray, line 710: It should be set true if the list was populated from default values, the point being that if it was fabricated then the list is not serialized. So what's most important is to skip serializing it when creating the flatbuffer instance (if fabricated is true). Then, this would go through the `None` case when deserializing.

Perhaps we can cook up some tests around the expected behavior.

jrray (changes requested):

crates/spk-schema/src/fb_converter.rs
```diff
@@ -0,0 +1,1358 @@
...
+    if let Some(compat) = fb_compat {
+        Compat::new_unchecked(compat)
+            .expect("A Compat in flatbuffer data should be a valid Compat when parsed")
+    } else {
+        // In this case, None, so nothing, stored as an fb_compat
+        // represents the default compat. This is different to the var
+        // opt compat method above.
+        Compat::default()
+    }
+}
```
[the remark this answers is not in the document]
dcookspi: Added to this ticket https://github.com/spkenv/spk/issues/1347

crates/spk-schema/src/v0/solver_package_spec.rs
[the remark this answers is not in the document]
dcookspi: I've renamed it to `IndexedPackage`.

crates/spk-schema/src/v0/solver_package_spec.rs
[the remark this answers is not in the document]
dcookspi: I've removed the buf field from the hashing. It should be the case that two different packages have different offsets within the same index.

crates/spk-schema/src/fb_converter.rs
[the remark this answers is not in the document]
dcookspi: I've updated the comment to reflect this.

crates/spk-schema/src/fb_converter.rs
```diff
@@ -0,0 +1,1358 @@
...
+                        let ident_name =
+                            unsafe { PkgNameBuf::from_string(fb_pkg_req.name().to_string()) };
+
+                        let components =
+                            fb_component_names_to_component_names_set(&fb_pkg_req.components());
+
+                        // TODO: this will be the next thing to get a
+                        // proper flatbuffer representation because it
+                        // is now showing up as the next significant
+                        // chunk on the large solve flamegraphs
```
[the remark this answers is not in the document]
dcookspi: This change has been added in Indexing PR6, https://github.com/spkenv/spk/pull/1344

crates/spk-proto/schema/spk.fbs
dcookspi: I've left it in.

crates/spk-proto/schema/spk.fbs
dcookspi: Updated the comment to reflect this.

crates/spk-proto/schema/spk.fbs
dcookspi: No, they are unsorted. I've added a comment saying that.

crates/spk-proto/schema/spk.fbs
dcookspi: I've added a field to store the index schema version number in. It isn't used in this PR because the top level index reading code is in the next change, PR4 . I'll update PR4 to make use of it.

crates/spk-proto/Cargo.toml
dcookspi: I've moved it do the workspace, and updated the reference in spfs too.

crates/spk-schema/src/fb_converter.rs
dcookspi: Yes, they are generated as constants with a number type, and lists of those constants. I've added debug_assert!s to these cases, and updated the comments.

crates/spk-schema/src/fb_converter.rs
dcookspi: I've added a check to the component embedded packages list generation for the index to only include non-fabricated lists.

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I've moved the offset check to the front, and the offset's addition to the hash to the front too.

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I've created an issue to collect the related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/indexed_package.rs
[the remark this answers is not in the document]
dcookspi: I've created an issue to collect the related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/solver_package_spec.rs
dcookspi: I've created an issue to collect the related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/solver_package_spec.rs
dcookspi: I've created an issue to collect the related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I've created an issue to collect the related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: It is an exact duplicate. The code could almost be the default implementation for this methond in the `Package` trait, except it relies on `check_satisfies_request()` which is in the `Satisfiy..,>` trait that the `Package` trait does not rely on - but both `PackageSpec` and `IndexedPackage` do.  I'll update the comment in this copy to mention this and add this to the issue that collects the trait refactoring related comments and discussions: https://github.com/spkenv/spk/issues/1348

crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -110,7 +110,7 @@ impl std::cmp::PartialEq for IndexedPackage {
     fn eq(&self, other: &Self) -> bool {
         // This deliberately ignores the buf field, offset field, and
         // all the caches.
-        self.build_ident == other.build_ident && self.offset == other.offset
+        self.offset == other.offset && self.build_ident == other.build_ident
```
dcookspi: This doesn't do what it was meant too, I'll be updating it again.

dcookspi: I've corrected it now.

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I tried swapping moka in for these ArcSwaps, but the current struct and use (storing 5 calculated fields per package on first access) doesn't lend itself to per IndexedPackage caches. The result, without rearchitecting, was a slowdown in the solve times. It added a few seconds to the larger solves, which makes sense given they access more than 10000 packages while solving.  

I've left these fields unchanged, but it could be something to come back to in a future change - either adding an architectural change like set of moka caches at the FlatbufferRepoIndex level and sharing them between all the IndexedPackages that index produces, or swapping out the ArcSwaps/Arcs for something like a std::sync::LazyLock instead.

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I tried moka, it wasn't a good fit for the current architecture and usage, see https://github.com/spkenv/spk/pull/1338#discussion_r3024024251 for details.

crates/spk-schema/src/spec.rs
```diff
@@ -630,9 +630,14 @@ impl Test for SpecTest {
 #[derive(Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd, Serialize)]
 #[serde(tag = "api")]
 #[enum_dispatch(Deprecate, DeprecateMut)]
+#[allow(clippy::large_enum_variant)]
```
jrray: Maybe I missed this before, but we don't want to allow this. You can avoid it by `Box`ing the `IndexedPackage`.

jrray (changes requested):

crates/spk-schema/src/spec.rs
dcookspi: I've done this. I had to box both of the entries because of the size differences between them and the Boxed object itself.

jrray (approved):

PR #1339, opened 2026-03-19
Indexes 4: Adds Indexes for SPK repositories

dcookspi:
_Note: for info on benefits of indexing for spk solves see https://github.com/spkenv/spk/pull/1340 (5 of 5). Maybe start there and work back down to this PR if you prefer to review PRs top down._

This adds a new `Indexed` repository type, `RepositoryIndex` traits,  a `RepoIndex` enum, and a flatbuffer index implementation in `FlatBufferRepoIndex`, along with some configuration. The indexing is designed to help the solvers. An index only contains information useful for solving. It does not contain enough information to do other things, such as build or test the packages.

The hierarchy of new types is roughly:
```
RepositoryHandle (existing spk repos)
   Indexed 
       RepoIndex
           FlatBufferRepoIndex  (makes SolverPackageSpecs, a v0 package alternate, see PR 3 of 5 below)
                flatbuffers schema (see PR 3 of 5 below)
```
This also updates the solver and repo based tests to include the new Indexed repo. This does not add command line tools or solver interaction for indexes. Those are in the last PR in this chain.

This is the change that adds indexes and index based packages to Spk repositories. 

This is 4 of 5 chained PRs for adding indexes to spk solves:
1. https://github.com/spkenv/spk/pull/1336
2. https://github.com/spkenv/spk/pull/1337
3. https://github.com/spkenv/spk/pull/1338
4. this PR
5. https://github.com/spkenv/spk/pull/1340
6. https://github.com/spkenv/spk/pull/1344
7. #1354
8. #1355
9. #1356


crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1174 @@
...
+            let fb_compat = compat_to_fb_compat(builder, &build_spec.compat());
+
+            spk_proto::BuildIndexArgs {
+                published_version: Some(fb_build_version),
+                build: Some(fb_build_id),
+                build_type: fb_build_type,
+                is_deprecated: build_spec.is_deprecated(),
+                compat: fb_compat,
+                // Does not store any more details because the package
+                // is deprecated.
```
dcookspi: This means that deprecated packages in the index cannot be safely used in current solves as things stand. Because they have no install requirements in the index, they will short-circuit any solve they are resolved in. This is a problem for solves that ask for a deprecated build specifically. However, the number of times this happens is vanishingly small in our experience. There are several options to address this: 

-add a `--deprecated` flag that disables index use if deprecated packages are wanted (it turns out this does not exist for solves at the moment) add a `--deprecated` flag that disables index use if deprecated packages are wanted (it turns out this does not exist for solves at the moment). These solves would lack the benefit of any index
-add a way of marking a specific request as allowing deprecated builds, like the --deprecated but for that request only. This could either disable index use, or we could split the solvers processing to first ask the repo if the build is deprecated, and then load the package data from the index or underlying repo as required.
-include the full solver require data for deprecated packages in the index, which increases the index size, loading time and generation time. The increases might be managable but the data would be unused for most solves.

This probably needs a separate discussion.

crates/spk-storage/src/storage/repository_index.rs
```diff
@@ -0,0 +1,98 @@
...
+        repo: &crate::RepositoryHandle,
+        package_version: &VersionIdent,
+    ) -> miette::Result<()>;
+}
+
+/// Type for wrapping different kinds of indexes
+#[derive(Debug, Clone)]
+pub enum RepoIndex {
+    Flat(FlatBufferRepoIndex),
+}
```
dcookspi: There were more kinds of index during development and testing. There may be more in future.

crates/spk-config/src/config.rs
```diff
@@ -87,6 +87,29 @@ pub struct Solver {
 
     /// Name of the solver whose output to show  when multiple solvers are being run.
     pub solver_to_show: String,
+
+    /// Whether to get the solver to use repository indexes instead of
```
```suggestion
    /// Whether to let the solver use repository indexes instead of
```

This sounds more natural IMO.

crates/spk-config/src/config.rs
```diff
@@ -87,6 +87,29 @@ pub struct Solver {
...
+
+    /// What kinds of index to use. Only applies if there is more than
+    /// one kinds of index available for the repositories. The default
+    /// is 'flatb', a flatbuffers file based index.
+    pub index_kind: String,
+
+    /// Whether to validate flatbuffers index data before using it.
+    /// Validating is safer but adds some overhead at the start of a
+    /// solve when using an index.
+    pub verify_flatbuffers_index_before_use: bool,
```
jrray: It's a bit awkward to both try to be generic about what index types are available but also have "top-level" configuration options that are specific to one index type.

For that reason I would suggest having a sub-struct for containing options pertaining to the flatbuffers index.

Calling it the flatbuffers index is a separate concern. Who's to say some other index wouldn't get created and also choose to use flatbuffers for the storage representation? In practice I don't see it being likely that we'll have competing index implementations, but it is nice to leave the possibility open.

crates/spk-solve/src/solvers/resolvo/spk_provider.rs
```diff
@@ -319,6 +333,41 @@ impl ResolvoPackageName {
...
+                            && build_ident == ident.build()
+                        {
+                            // Allow it to continue because, even though it is deprecated,
+                            // the current request is for a specific build and this
+                            // build matches the request.
+                        } else {
+                            // Exclude this deprecated build.
+                            let reason = provider
+                                .pool
+                                .intern_string(format!("{} is deprecated", ident));
```
jrray, line 365:
```suggestion
                                .intern_string(format!("{ident} is deprecated"));
```

crates/spk-solve/src/solvers/solver_test.rs
```diff
@@ -175,11 +194,14 @@ async fn test_solver_no_requests(#[case] mut solver: SolverImpl) {
...
+#[case::step(step_solver(), false)]
+#[case::step_indexed(step_solver(), true)]
+#[case::resolvo(resolvo_solver(), false)]
+#[case::resolvo_indexed(resolvo_solver(), true)]
 #[tokio::test]
 async fn test_solver_package_with_no_recipe(
     #[case] mut solver: SolverImpl,
+    #[case] use_index: bool,
     random_build_id: BuildId,
 ) {
```
jrray:
```suggestion
#[case::step(step_solver())]
#[case::resolvo(resolvo_solver())]
#[tokio::test]
async fn test_solver_package_with_no_recipe(
#[case] mut solver: SolverImpl,
    #[values(true, false)] use_index: bool,
random_build_id: BuildId,
) {
```

I'd prefer this pattern for simplicity / extensibility. I was thinking we'd have a declarative macro to help with removing boilerplate on every test.

crates/spk-solve/src/solvers/solver_test.rs
```diff
@@ -197,6 +219,7 @@ async fn test_solver_package_with_no_recipe(
     repo.publish_package(&spec.into(), &components)
         .await
         .unwrap();
+    let repo = wrap_repo_for_test(repo, use_index).await;
```
jrray, line 220: It's not ideal that every test has to remember to call this, though I guess there would be a lint that `use_index` is unused if that's the case.

crates/spk-solve/src/solvers/solver_test.rs
```diff
@@ -2797,6 +3106,7 @@ async fn test_solver_component_embedded_component_requirements(
 #[tokio::test]
 async fn test_solver_component_embedded_multiple_versions(
     #[values(step_solver(), resolvo_solver())] mut solver: SolverImpl,
+    #[values(false, true)] use_index: bool,
```
jrray, line 3029: Ayyy...

You could argue this is better than defining cases for the solver flavors.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1174 @@
...
+impl FlatBufferRepoIndex {
+    /// Verify the flatbuffer data is a valid repository index
+    fn check_fb_index(&self) -> Result<spk_proto::RepositoryIndex<'_>> {
+        let verifier_opts = flatbuffers::VerifierOptions {
+            // The default number of tables seems to be 1 million, and
+            // that isn't enough for the current index. Tried 100
+            // million and that didn't error, but this is not a great
+            // solution long term, especially once VersionFilters are
+            // converted from strings.
+            max_tables: MAX_FLATBUFFER_TABLES,
```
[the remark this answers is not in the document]
jrray: Along the lines of the idea to do the verification after writing a new index file, can we find out how many tables were used and print a warning if it is some percentage of the maximum?

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1209 @@
...
+                    return Err(Error::String(format!(
+                        "Unable to remove existing index file '{}' due to: {err}",
+                        filepath.display()
+                    )));
+                }
+            };
+        }
+
+        // Create the new index file with the correct permissions
+        if let Err(err) = tokio::fs::write(&filepath, builder.finished_data()).await {
```
jrray: We need to write to a temp file first and rename if successful.

crates/spk-storage/src/storage/flatbuffer_index_test.rs
```diff
@@ -0,0 +1,512 @@
...
+    make_build,
+    //make_build_and_components,
+    make_package,
+    //pinned_request,
+};
+
+use crate::{IndexedRepository, RepoWalkerBuilder, RepoWalkerItem, RepositoryHandle};
+
+// A copy of the one in spk-solve-macros because using it directly
+// here cases an import loop
```
jrray, line 41:
```suggestion
// here causes an import loop
```

crates/spk-storage/src/storage/handle.rs
```diff
@@ -42,11 +43,16 @@ impl RepositoryHandle {
         matches!(self, Self::Runtime(_))
     }
 
+    pub fn is_indexed(&self) -> bool {
```
jrray: If you add `Variantly` to the derive list you get this method for free.

jrray (changes requested):

crates/spk-config/src/config.rs
```diff
@@ -87,6 +87,36 @@ pub struct Solver {
 
     /// Name of the solver whose output to show  when multiple solvers are being run.
     pub solver_to_show: String,
+
+    /// Whether to get the solver to use repository indexes instead of
+    /// the repository directly.
+    pub use_indexes: bool,
+
```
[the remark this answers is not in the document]
dcookspi: Docs for these have been added to PR 5 https://github.com/spkenv/spk/pull/1340

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: This comment is out of date now. I tested adding all the deprecated data to the index, and while it increases the index side, it doesn't impact the solve times (see PR5 https://github.com/spkenv/spk/pull/1340 for some times).

- [x] Update the comment in the code

dcookspi:
> I can't replicate the test failures locally at the moment. I'll dig into them further tomororw.

These have been fixed now.

crates/spk-storage/src/storage/flatbuffer_index.rs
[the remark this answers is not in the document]
dcookspi: I think we can. I've added a log message to show this info. I'll add a check for a warning in PR5 (#1344) when retesting the command line tools.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated it to do this. It still removes the old index file before renaming the temp file over it.

crates/spk-storage/src/storage/flatbuffer_index.rs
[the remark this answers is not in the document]
dcookspi: It turned out we can't easily get this information from the flatbuffers api. So I've taken the log message out. We'd have to count the tables ourselves as the index is generated, using info about which rust objects are represented in tables in the index schema.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1174 @@
...
+                published_version: Some(fb_build_version),
+                build: Some(fb_build_id),
+                build_type: fb_build_type,
+                is_deprecated: build_spec.is_deprecated(),
+                compat: fb_compat,
+                // Does not store any more details because the package
+                // is deprecated.
+                // TODO: might want to add something to load them from
+                // underlying repo if required, such as when the
+                // --deprecate flag is set.
```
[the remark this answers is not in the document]
dcookspi: I've removed this comment because the index holds the deprecated packages data now.

crates/spk-storage/src/storage/flatbuffer_index_test.rs
```diff
@@ -0,0 +1,499 @@
...
+    assert_repo_and_index_have_same_packages(repo1, indexed_repo1).await;
+
+    println!("Repo2's");
+
+    let indexed_repo2 = index_for_test(repo2.clone()).await;
+
+    assert_repo_and_index_have_same_packages(repo2, indexed_repo2).await;
+}
+
+// TODO: add rest of solves sample repos to this as tests
```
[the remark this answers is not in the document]
dcookspi: I've made an issue to streamline the test repo setup in future: https://github.com/spkenv/spk/issues/1352

crates/spk-config/src/config.rs

crates/spk-solve/src/solvers/resolvo/spk_provider.rs
dcookspi: I've updated this.

crates/spk-solve/src/solvers/solver_test.rs
dcookspi: I've switched all the tests over to use values rather than be test cases.

crates/spk-storage/src/storage/flatbuffer_index_test.rs
dcookspi: I've updated this.

crates/spk-solve/src/solvers/solver_test.rs
dcookspi: I've updated the rest of the tests to use values.

crates/spk-storage/src/storage/handle.rs
dcookspi: I've added the `Variantly` derive to the `RepositoryHandle` enum and removed the other is_* methods.

crates/spk-config/src/config.rs
dcookspi: I took the 'flatbuffers' out of that setting name so we can treat it as more generic.  This lets it used to enable/disable veriftication checking for any indexes, even it's a no-op for a particular index type.  

There are three index related settings in the solver settings config here. They could all be moved to an indexes sub-section of the solver, if that would help with the separation?

crates/spk-solve/src/solvers/solver_test.rs
dcookspi: I've added this comment to repo/solver test setup issue: https://github.com/spkenv/spk/issues/1352

crates/spk-config/src/config.rs
jrray: It's more about not have settings at a top level that aren't applicable in some configurations. Having a sub-struct to group the index settings could help organize, if those settings are applicable to any/all index implementation. If any are specific to a particular index implementation then that implementation should have its own sub-[sub-]struct.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1259 @@
...
+    /// remove the old index file if any, and move the new temp file
+    /// into its place.
+    async fn save_index<'a>(
+        repo: &crate::RepositoryHandle,
+        builder: &flatbuffers::FlatBufferBuilder<'a>,
+    ) -> Result<PathBuf> {
+        let name = repo.name();
+
+        let filepath = Self::repo_index_location(repo).await?;
+        let temp_file = PathBuf::from(format!("{}_being_generated", filepath.display()));
```
jrray: Please use a unique filename via uuid/ulid/tempfile but you're on the right track. The temp file should be in the same directory as the final file.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1259 @@
...
+        tracing::debug!("Index temp file: {}", temp_file.display());
+
+        // Create the new index in a temp file with the correct permissions
+        if let Err(err) = tokio::fs::write(&temp_file, builder.finished_data()).await {
+            return Err(Error::IndexWriteError(
+                name.to_string(),
+                temp_file.display().to_string(),
+                err,
+            ));
+        } else {
```
jrray, line 692: All the error handlers at this point forward should attempt to delete the temp file.

jrray (changes requested):

crates/spk-config/src/config.rs
dcookspi: I've pulled out some of the index config into a struct for an Index. It is used with the solver and the repository config structs. Each now has a flag to uses indexes and a sub-section for index config settings (currently, kind and verfiy before use). These config structures are not used much in this PR, but I can test and adjust them in PR5 (https://github.com/spkenv/spk/pull/1340) as needed.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've added a ulid to the temp file name.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've pulled out the index file removal code into a function and used it for the temp files in each of the error handling points.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -0,0 +1,1281 @@
...
+            // Ensure the file is readable and writable by everyone
+            #[cfg(unix)]
+            match tokio::fs::set_permissions(&temp_file, Permissions::from_mode(0o666)).await {
+                Err(err) => {
+                    // Clean up the temp file after the error
+                    remove_index_file(&temp_file).await?;
```
jrray: Okay but if removing the temp file fails the `?` usage here will mask the original error. Any errors in cleanup could be logged but the caller should get the first error as the return value.

jrray (changes requested):

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've update all the places like this to log and return the first error.

jrray (approved):

PR #1349, opened 2026-03-30
Add suppress for strong var inheritance

### LINKED ISSUE #1324: Support opting out of strong inheritance vars
Our use case is how we have our python package inject an install requirement into downstream packages:

```yaml
build:
  options:
  - var: abi/cp311
    choices:
    - cp311
    - cp311d
    inheritance: Strong
    description: The Python ABI flavor to use.
```

This helps prevent accidentally forgetting to add this to downstream packages in the general case but sometimes a package only uses python as a tool as part of the build script but has no runtime requirement on python.

What we lack is a way for those packages to prevent inheriting the var.

Attempting to declare the var in the downstream package doesn't work:

```yaml
install:
  requirements:
    # block inheriting python.abi from python
    - var: python.abi/
```

    ERROR   × Error: Cannot insert requirement: one already exists and only pkg requests can be merged: var: python.abi/ + var: python.abi/cp311

We could support this syntax but it is too easy to do something like this unintentionally. It would be more self-documenting to add a new attribute specifically for this opt-out purpose:

```yaml
install:
  requirements:
    # block inheriting python.abi from python
    - var: python.abi
      suppress: "this package doesn't use python at runtime"
```

It could be a boolean field (or not care what the value is) but here I used the value position as a place to document why it is there.

jrray:
Allow downstream packages to opt out of inheriting strongly-inherited variables by adding a suppress
directive to install requirements:

```yaml
  install:
    requirements:
      - var: base.inherit-me
        suppress: "reason for suppressing"
```

The suppress directive prevents the variable from
being added to the built package's runtime
requirements and tells the InheritRequirements
validator to skip checking for it.

Closes #1324.

### DIFF
--- a/crates/spk-build/src/build/binary.rs
+++ b/crates/spk-build/src/build/binary.rs
@@ -392,6 +392,7 @@ where
             package,
             variant: full_variant,
             environment_filesystem,
+            suppressed_requirements: self.recipe.suppressed_requirements(),
         };
         let mut report = BuildReport {
             setup,
--- a/crates/spk-build/src/validation/inherit_requirements.rs
+++ b/crates/spk-build/src/validation/inherit_requirements.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/spkenv/spk
 
+use spk_schema::ident::RequestWithOptions;
 use spk_schema::name::PkgNameBuf;
 use spk_schema::validation::{
     ValidationMatcherDiscriminants,
@@ -88,6 +89,11 @@ impl super::Validator for InheritRequirementsValidator<'_> {
                     .spec
                     .downstream_runtime_requirements([&component]);
                 for request in downstream_runtime.iter() {
+                    if let RequestWithOptions::Var(var) = request
+                        && setup.suppressed_requirements.contains(&var.var)
+                    {
+                        continue;
+                    }
                     let status = match (self.kind, runtime_requirements.contains_request(request)) {
                         (RuleKind::Allow, Compatibility::Compatible)
                         | (RuleKind::Allow, Compatibility::Incompatible(_))
--- a/crates/spk-build/src/validation/inherit_requirements.rs
+++ b/crates/spk-build/src/validation/inherit_requirements.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/spkenv/spk
 
+use spk_schema::ident::RequestWithOptions;
 use spk_schema::name::PkgNameBuf;
 use spk_schema::validation::{
     ValidationMatcherDiscriminants,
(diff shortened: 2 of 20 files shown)


crates/spk-build/src/build/binary.rs
jrray, line 390: I think when this `BuildSetupReport` type was conceived, at the time there was little difference between the recipe and package types, so the fact this type only carries the (built) package didn't matter as much. But these days it would likely make sense to have both the recipe and the package here. There are fields in the recipe that don't get added to the package.

This PR currently takes a more minimal approach of adding `suppressed_requirements` as a field, but it could stick the whole recipe in here so the validation rules have access to it.

crates/spk-build/src/validation/inherit_requirements.rs
```diff
@@ -88,6 +89,11 @@ impl super::Validator for InheritRequirementsValidator<'_> {
                     .spec
                     .downstream_runtime_requirements([component]);
                 for request in downstream_runtime.iter() {
+                    if let RequestWithOptions::Var(var) = request
```
jrray, line 92: This was the main implementation hurdle, making it so if the var is suppressed it doesn't then violate the validation rules.

crates/spk-schema/crates/foundation/src/ident/pinnable_request.rs
```diff
@@ -195,17 +195,41 @@ impl<'de> Deserialize<'de> for PinPolicy {
...
 /// Represents a constraint added to a resolved environment.
 #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Variantly)]
-#[cfg_attr(feature = "parsedbuf-serde", derive(Serialize), serde(untagged))]
 pub enum PinnableRequest {
     Pkg(PkgRequest),
     Var(VarRequest<PinnableValue>),
+    /// Prevent the package from inheriting the var named.
+    /// Contains the var name and an optional comment explaining
+    /// why the suppression is needed.
+    Suppress(OptNameBuf, String),
```
jrray, line 204: I elected to add a 3rd arm here to not overload the `VarRequest` type and have that impact many more areas of the codebase. This feature should conceptually stay within the realm of recipes and building packages and not extend into the built package or solver territories.

docs/ref/api/v0/package.md
```diff
@@ -386,6 +386,7 @@ A build option can be one of [VariableRequest](#variablerequest), or [PackageReq
 | var                 | _str_  | The requested value of a package build variable in the form`name=value`, this can reference a specific package or the global variable (eg `debug=on`, or `python.abi=cp37`)                  |
 | fromBuildEnv        | _bool_ | If true, replace the requested value of this variable with the value used in the build environment                                                                                           |
 | ifPresentInBuildEnv | _bool_ | Either true or false; if true, then `fromBuildEnv` only applies if the variable was present in the build environment. This allows different variants to have different runtime requirements. |
+| suppress            | _str_  | If set, prevents this variable from being inherited from upstream packages with Strong inheritance. The value serves as documentation for why the suppression is needed.                     |
```
rydrman, line 389: something about not compatible with any other fields?

rydrman (approved):

PR #1344, opened 2026-03-28
Indexes 6: Changes `version_filter` field in index schema

dcookspi:
This changes the `version_filter` field in the index schema to a flatbuffers list and table representation, instead of a string. This allows the version filter rules (VersionRange entries) to be extracted from the index without string parsing.

The resulting index is slightly larger at 122 MB (see Indexes PR 5: https://github.com/spkenv/spk/pull/1340). Index generation time is about the same. But the solve times improve for the larger solves (marked with *):
```
Requests        | Solution size | Num.    | Solve time  |  Indexed solve time |  Solve time now
                | (# packages)  | Retries | (seconds)   |  (seconds)          |  (seconds)  
------------------------------------------------------------------------------------------------
python          |        4      |    1    |     0.17    |   0.03              |  0.04
boost-python    |        8      |    1    |     0.31    |   0.05              |  0.07
python-torch    |       37      |    2    |     0.58    |   0.15              |  0.14
widget toolset  |       60      |    2    |     3.44    |   0.48              |  0.38
katana toolset  |      181      |   10    |    18.32    |   2.97              |  2.14
nuke toolset    |      280      |   12    |    24.32    |   4.55              |  1.70  (*)
houdini toolset |      211      |   19    |    37.24    |   6.58              |  3.78  (*)
maya toolset    |      403      |   20    |    59.50    |   9.52              |  5.30  (*)
```

This is the 6th of the chained PRs for adding indexes to spk solves:
1. https://github.com/spkenv/spk/pull/1336
2. https://github.com/spkenv/spk/pull/1337
3. https://github.com/spkenv/spk/pull/1338
4. https://github.com/spkenv/spk/pull/1339
5. https://github.com/spkenv/spk/pull/1340
6. this PR
7. #1354
8. #1355
9. #1356

PR #1340, opened 2026-03-20
Indexes 5: Adds `spk repo index` subcommand for index generation and updates

dcookspi:
This adds a new `repo index` subcommand to spk for index generation and updates. It adds the `--use-indexes` and `--no-indexes` flags for repository index usage. This also updates `resolvo` solver to get global variables data from an indexed repository. This allows `resolvo` to solve without needing to restart its solves.

### Indexing
The index is designed to help the solvers with solve times. It doesn't contain enough data to help with other spk operations like building and testing a package. 

Indexing can be enabled or disabled in the spk config file. If indexing is enabled, you have to generate an index, with `spk repo index`, prior to trying to use it. They are not generated on the fly (outside of automated tests).

To generate an index (for the origin):
- `spk repo index --disable-repo local`

To update an existing index, e.g. after a new python package was published:
- `spk repo index --disable-repo local --update python`

The flatbuffer index data is stored in a file in the underlying spfs repo in a `index/spk/` sub-directory.

If index use is enabled in the config file, it can be disabled with the `--no-indexes` command line flag. If index use is disabled by default, it can be enbled with the `--use-indexes` flag. If index use is enabled, but no index has been generated, spk will fallback to using the underlying repo directly (it acts as it would before this change).

### Speed Diferences
Generating the index file on our repo (sizes below) takes about 2 minutes. Updating a package in an existing index, such as after a new build is published, takes a few seconds.

#### Sample solver time improvements using this indexing
The numbers come from this setup:
- an origin repo that has 2245 packages, 23540 versions, 82517 builds (11 erroring, about 30% deprecated), and 141 global vars
- the index loads in ~0.0003 seconds unverified, or ~0.2 seconds verified, and is about 76 MB on disk with trimmed down deprecated builds (107 MB with full deprecated builds)
- these times are from a rough average of 3-4 runs with index verification disabled 
- a "toolset" below is a set of requests for the named DCC and our typical in-house plugins and tools
```
Requests        | Solution size | Num.    | Solve time  |  Indexed solve time, no retries
                | (# packages)  | Retries | (seconds)   |  (seconds)
-----------------------------------------------------------------------------------------
python          |        4      |    1    |     0.17    |   0.03 
boost-python    |        8      |    1    |     0.31    |   0.05 
python-torch    |       37      |    2    |     0.58    |   0.15 
widget toolset  |       60      |    2    |     3.44    |   0.48 
katana toolset  |      181      |   10    |    18.32    |   2.97 
[...]

PR #1354, opened 2026-04-16
Indexes 7: Adds a lock file around index generation and updates

dcookspi:
This adds a lock file around index generation and update processing. The lock for the index file is created, via a `FileLock` object, before the underlying repo is read and kept until the index has been generated and written out. The same duration of locking is done for updating the index. 

A `FileLock` has configurable settings for the maximum number of tries to get the lock before giving up, and the amound of time to sleep between tries, as well as the kind of file being locked (used in log messages).  It also contains sentry integration when the sentry feature is enabled.

This is the 7th of the chained PRs for adding indexes to spk solves:

1. #1336
2. #1337
3. #1338
4. #1339
5. #1340
6. #1344
7. this PR
8. #1355
9. #1356
10. #1363

### DIFF
--- a/crates/spk-cli/cmd-repo/src/cmd_repo.rs
+++ b/crates/spk-cli/cmd-repo/src/cmd_repo.rs
@@ -135,22 +135,36 @@ impl RepoCommand {
 
                     // Load the current index for this repo now
                     let mut was_full_index = String::from("");
-                    match FlatBufferRepoIndex::from_repo_file(&repo_to_index).await {
+                    let result = match FlatBufferRepoIndex::from_repo_file(&repo_to_index).await {
                         Ok(current_index) => {
-                            current_index
-                                .update_packages(&repo_to_index, &idents)
-                                .await?
+                            current_index.update_packages(&repo_to_index, &idents).await
                         }
                         Err(err) => {
                             // There isn't an existing index, so generate one from scratch that
                             // will also include the update package version.
                             tracing::warn!("Failed to load flatbuffer index: {err}");
                             tracing::warn!("No current index to update. Creating a full index ...");
-                            FlatBufferRepoIndex::index_repo(&repos).await?;
                             was_full_index =
-                                " [no previous index, so a full index was created]".to_string()
+                                " [no previous index, so a full index was created]".to_string();
+                            FlatBufferRepoIndex::index_repo(&repos).await
                         }
                     };
+
+                    if result.is_err() {
+                        // Need to keep these if-statements separate
+                        // to allow for two different error handling cases.
+                        #[allow(clippy::collapsible_if)]
+                        if let Some(err) = result.err() {
+                            if let Some(spk_storage::Error::UnableToGetWriteLockError(..)) =
+                                err.root_cause().downcast_ref::<spk_storage::Error>()
+                            {
+                                tracing::error!("{err}");
+                                // A distinct exit code when unable to lock the index file.
+                                return Ok(3);
+                            }
+                            return Err(err);
+                        }
+                    }
 
                     tracing::info!(
                         "Index update for '{}' in '{}' repo completed in: {} secs{was_full_index}",
--- a/crates/spk-cli/cmd-repo/src/cmd_repo.rs
+++ b/crates/spk-cli/cmd-repo/src/cmd_repo.rs
@@ -135,25 +135,36 @@ impl RepoCommand {
 
                     // Load the current index for this repo now
                     let mut was_full_index = String::from("");
-                    match FlatBufferRepoIndex::from_repo_file(&repo_to_index).await {
+                    let result = match FlatBufferRepoIndex::from_repo_file(&repo_to_index).await {
                         Ok(current_index) => {
-                            current_index
-                                .update_packages(&repo_to_index, &idents)
-                                .await?
+                            current_index.update_packages(&repo_to_index, &idents).await
                         }
                         Err(storage::Error::IndexOpenError(err)) => {
                             // There isn't an existing index, so generate one from scratch that
                             // will also include the update package version.
                             tracing::warn!("Failed to load flatbuffer index: {err}");
                             tracing::warn!("No current index to update. Creating a full index ...");
-                            FlatBufferRepoIndex::index_repo(&repos).await?;
                             was_full_index =
-                                " [no previous index, so a full index was created]".to_string()
+                                " [no previous index, so a full index was created]".to_string();
+                            FlatBufferRepoIndex::index_repo(&repos).await
                         }
                         Err(err) => {
                             return Err(err.into());
                         }
                     };
+
+                    // Need to keep these if-statements separate
+                    // to allow for two different error handling cases.
+                    if let Some(err) = result.err() {
+                        if let Some(spk_storage::Error::UnableToGetWriteLockError(..)) =
+                            err.root_cause().downcast_ref::<spk_storage::Error>()
+                        {
+                            tracing::error!("{err}");
+                            // A distinct exit code when unable to lock the index file.
+                            return Ok(3);
+                        }
+                        return Err(err);
+                    }
 
                     tracing::info!(
                         "Index update for '{}' in '{}' repo completed in: {} secs{was_full_index}",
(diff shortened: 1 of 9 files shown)


crates/spk-cli/cmd-repo/src/cmd_repo.rs
```diff
@@ -135,23 +135,37 @@ impl RepoCommand {
...
-                                " [no previous index, so a full index was created]".to_string()
+                                " [no previous index, so a full index was created]".to_string();
+                            FlatBufferRepoIndex::index_repo(&repos).await
                         }
                     };
 
+                    if result.is_err() {
+                        // Need to keep these if-statements separate
+                        // to allow for two different error handling cases.
+                        #[allow(clippy::collapsible_if)]
```
jrray: Can you lose the outer `if`?

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +83,288 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+            match tokio::fs::OpenOptions::new()
+                .create_new(true)
+                .write(true)
+                .open(&lock_file)
+                .await
+            {
+                Ok(_file) => {
+                    // Write any given contents to the lock file as well
+                    if let Some(ref data) = contents
+                        && let Err(err) = tokio::fs::write(&lock_file, data).await
```
jrray:
```suggestion
                Ok(mut file) => {
                    // Write any given contents to the lock file as well
                    if let Some(ref data) = contents
                        && let Err(err) = tokio::io::AsyncWriteExt::write_all(&mut file, data).await
```

Use the file handle already opened.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +83,288 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+                    };
+                }
+            }
+        }
+
+        // Unable to get a lock at this point. Read the existing lock
+        // file for the lock data and add it to the returned error.
+        use std::io::Read;
+        let mut lock_data = String::new();
+        let held_for = match std::fs::File::open(&lock_file) {
```
jrray: It's weird to switch to synchronous file apis here in an async function that was using async file apis earlier.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +83,288 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+            sleep_seconds,
+            lock_data,
+            held_for,
+        ))
+    }
+
+    /// Unlock this file lock by removing the lock file
+    ///
+    /// # Safety:
+    /// File locks are used to ensure that only one process is writing
```
jrray:
```suggestion
    /// # Safety
    ///
    /// File locks are used to ensure that only one process is writing
```

The style convention we're using _almost_ consistently throughout the project.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +83,288 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+                    }
+
+                    let unlock_error = Error::UnableToRemoveWriteLockError(
+                        self.label.clone(),
+                        self.lock_file.display().to_string(),
+                        err.to_string(),
+                    );
+
+                    return Err(unlock_error);
+                }
```
jrray, line 356:
```suggestion
                } else {
                    self.has_been_deleted = true;
                }
```

If the file didn't exist, then it had been deleted. Or does this variable mean to say `was_deleted_by_us`? Since this var is used to decide to attempt to delete the file in the drop handler then it should be set true here too, IMO.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +83,288 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+        Ok(())
+    }
+}
+
+impl Drop for FileLock {
+    fn drop(&mut self) {
+        if !self.has_been_deleted
+            && let Err(err) = unsafe { self.unlock() }
+        {
+            // Can only the problem here. The unlock() method will
```
jrray:
```suggestion
            // Can only <word> the problem here. The unlock() method will
```

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -1030,11 +1313,14 @@ impl RepositoryIndexMut for FlatBufferRepoIndex {
             )
             .into());
         }
+        let repo = &repos[0].1;
+
+        // Lock the index file until the data has been updated
+        let mut _lock = lock_index_file(repo).await?;
```
jrray: It will reduce programming bugs if the return value can be turned into a lock guard that is required as a parameter to the function call(s) that write to the index.

jrray (changes requested):

crates/spk-cli/cmd-repo/src/cmd_repo.rs
dcookspi: Yes, I've removed that,

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated it.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: Added this in.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've added it as a parameter

crates/spk-cli/cmd-repo/src/cmd_repo.rs
```diff
@@ -135,23 +135,35 @@ impl RepoCommand {
...
-                                " [no previous index, so a full index was created]".to_string()
+                                " [no previous index, so a full index was created]".to_string();
+                            FlatBufferRepoIndex::index_repo(&repos).await
                         }
                     };
 
+                    // Need to keep these if-statements separate
+                    // to allow for two different error handling cases.
+                    #[allow(clippy::collapsible_if)]
+                    if let Some(err) = result.err() {
```
jrray, line 158:
```suggestion
                    if let Some(err) = result.err() {
```

This can be removed, it was the unnecessary outer `if` causing the lint.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -82,6 +84,299 @@ async fn remove_index_file(filepath: &PathBuf) -> Result<()> {
...
+            max_tries,
+            sleep_seconds,
+            lock_data,
+            held_for,
+        ))
+    }
+
+    /// Unlock this file lock by removing the lock file
+    ///
+    /// # Safety:
```
jrray:
```suggestion
    /// # Safety
```

jrray (changes requested):

jrray:
Is it intentional that this PR has index-5 as the merge target and not index-6?

dcookspi:
> Is it intentional that this PR has index-5 as the merge target and not index-6?

It has `index-5-zplus-1-...` as the target, which is really number 6, but it wasn't renamed before its PR was made.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-cli/cmd-repo/src/cmd_repo.rs
dcookspi: I've removed that now.

jrray (approved):

PR #1356, opened 2026-04-20
Indexes 9: Adds messaging on package events to kafka

dcookspi:
This adds configurable messaging support to `spk` so it can send messages after events.

This adds support for sending message to kafka when these events occur:
- publishing a package
- modifiying a package
- removing a package

This allows a site to have an external messaging system listen for such messages and take appropriate action, e.g. updating the index on a repo after a new package has been published.

This is the 9th of the chained PRs for adding indexes to spk solves:

1. #1336
2. #1337
3. #1338
4. #1339
5. #1340
6. #1344
7. #1354
8. #1355 
9. this PR
10. #1363


crates/spk-config/src/config.rs
```diff
@@ -196,6 +196,30 @@ pub struct HostOptions {
...
+/// Configuration for using Kafka as a message channel.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct KafkaChannel {
+    /// List of brokers to connect to, specified as "hostname:port"
+    pub brokers: Vec<String>,
+    /// Topic name to push package updates to
+    pub package_updates_topic_name: Option<String>,
+    /// Message timeout in milliseconds, defaults to 5000 ms (5 seconds)
+    #[serde(default = "default_kafka_timeout_ms")]
+    pub timeout_ms: u64,
```
```suggestion
    /// Message timeout in milliseconds, defaults to 5000 ms (5 seconds)
    #[serde(default = "default_kafka_message_timeout_ms")]
    pub message_timeout_ms: u64,
```

I suggest mirroring the name of the kafka property as a convention to make it more plain what this setting maps to on the kafka side.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -0,0 +1,74 @@
...
+                .key(ident.name().as_str())
+                .payload(
+                    &json!({
+                        "event": event.to_string(),
+                        "repo": to.to_string(),
+                        "package": ident.to_string(),
+                    })
+                    .to_string(),
+                ),
+            Duration::from_secs(0),
```
jrray: This timeout is for waiting for room in the producer queue, it should be willing to wait some small amount of time as a form of back-pressure. Better to wait a little than fail aggressively.

cspell.json
```diff
@@ -140,6 +141,7 @@
     "depvar",
     "depver",
     "Deque",
+    "derserializer",
```
jrray: Why?

jrray (changes requested):

crates/spk-config/src/config.rs
dcookspi: I've updated this, and I'll do a pass over the next PR (#1363)  to ensure its config settings align as well.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've increased this to 4 seconds and made a constant for it for this PR. The next PR (#1363) adds another producer  and its send call will need updating too, and at that point I think I'll make the value into a configuration setting.

cspell.json
dcookspi: I've taken the extra r out.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -0,0 +1,81 @@
...
+use spk_config::KafkaChannel;
+use spk_schema::BuildIdent;
+use tokio::sync::Mutex;
+
+use crate::storage::messaging::PackageEvent;
+use crate::{Error, Result};
+
+/// The queue timeout for a producer sending a message to a queue.
+/// This determines how to long retry for if the producer queue is full.
+const PRODUCER_QUEUE_TIMEOUT_S: u64 = 4;
```
```suggestion
const PRODUCER_QUEUE_TIMEOUT: Duration = Duration::from_secs(4);
```

Safety via types.

jrray (changes requested):

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated this, although it goes away in the next PR (#1363), when it turned into a config value.

jrray (approved):

PR #1359, opened 2026-04-24
Fix required var not propagated to solver

jrray:
When a recipe declares a namespaced var option like `mylib.namespace_style/major_minor`, the value was added as a standalone VarRequest but not attached to the corresponding PkgRequestWithOptions. The solver needs the var in the package request's options map (as Complete) to satisfy a required var constraint.

Collect namespaced var options in get_build_requirements() and attach them to their corresponding package requests. Also fix the checks solver's impossible request pre-check to translate request options into build options on the dummy recipe so the combining logic preserves them as Complete.


dcookspi (approved):

jrray:
I misclicked "merge" when trying to click the dropdown to rebase, but rebased locally to back out the merge of main into this PR.

PR #1360, opened 2026-04-24
Add #[spfstest] macro to remove spfs run wrapper

jrray:
## Summary

- Add `#[spfstest]` proc macro that enables tests requiring an spfs runtime to self-exec into `spfs run - --` when not already in a runtime
- Annotate all 94 test functions that use `spfs_runtime()` with `#[spfstest]`
- Remove `spfs run - --` wrapper from `make test`
- `cargo nextest run` now works, reducing wall-clock test time from **5m30s to 1m10s** (~4.7x speedup)

## How it works

When a test marked `#[spfstest]` detects it is **not** inside an spfs runtime (`$SPFS_RUNTIME` not set), it re-launches the test binary inside `spfs run - --` with the original command-line arguments forwarded. When it **is** inside a runtime, the test runs directly.

Safety mechanisms:
- `AtomicBool` prevents redundant re-execs when multiple `#[spfstest]` tests share a process (`cargo test`)
- `__SPFSTEST_DEPTH` env var prevents infinite recursion if `spfs run` fails to set up a runtime
- Async and sync variants chosen automatically by the macro based on function signature

## When to use `#[spfstest]`

Add `#[spfstest]` to any test function that requires an active spfs runtime — typically any test that calls `spfs_runtime().await` or `spfs::active_runtime()`. Place it **before** `#[rstest]` and `#[tokio::test]`:

```rust
use spfstest::spfstest;

#[spfstest]
#[rstest]
#[tokio::test]
async fn test_something(tmpdir: tempfile::TempDir) {
    let rt = spfs_runtime().await;
    // ...
}
```

Tests that don't need a runtime should **not** use this attribute — it adds subprocess overhead.

## Nextest support

`cargo nextest run` now works out of the box:

```
$ time cargo nextest run --workspace --features server,spfs/server
Summary [  68.561s] 1419 tests run: 1419 passed, 0 skipped

real    1m9.748s
```

Compared to `cargo test`:

```
$ time cargo test --workspace --features server,spfs/server
real    5m30.356s
```

Nextest runs each test as a separate process, which is why the `#[spfstest]` re-exec approach is a natural fit — each test independently enters its own spfs runtime.

## New crates

| Crate | Type | Purpose |
|-------|------|---------|
| `spfstest-macro` | proc-macro | `#[spfstest]` attribute macro |
| `spfstest` | library | Re-exports macro + runtime re-exec helpers |

## Test plan

- [x] `cargo test --workspace --features server,spfs/server` — 1440 tests pass (no `spfs run` wrapper)
- [x] `cargo nextest run --workspace --features server,spfs/server` — 1419 tests pass
- [x] `spfs run - -- cargo test --workspace --features server,spfs/server` — tests still pass when already in a runtime (direct path, no re-exec)
- [x] Non-annotated tests unaffected


jrray:
## Note on test count discrepancy

`cargo test` reports 1440 tests while `cargo nextest run` reports 1419. The 21 extra in `cargo test` are double-counted due to the re-exec mechanism.

When `cargo test` runs a binary, all tests in that binary execute sequentially in one process. Tests that run *before* the first `#[spfstest]` test complete normally. Then the first `#[spfstest]` test triggers re-exec via `spfs run - --`, which runs the entire binary again inside the spfs runtime — replaying all tests, including the non-spfstest ones that already ran. After the subprocess finishes, `process::exit(0)` terminates the outer process.

So any non-spfstest tests that happened to run before the first spfstest test in the same binary get counted twice (once in the outer process, once in the subprocess). The 21 extra are exactly these tests.

With nextest this doesn't happen because each test is its own process — each spfstest test re-execs only itself.

rydrman (approved):
Love it!

PR #1361, opened 2026-04-30
Scope namespaced var requirements to their package

jrray:
## Summary

- Fixes a bug in the step-based solvers (cli, checks) where `VarRequirementsValidator` matched options by **base name**, so a namespaced var install requirement such as `demo.samenameaspkg/true` collided with an unrelated option sharing that base name — e.g. the version of a package named `samenameaspkg`, stored in the option map as `samenameaspkg=~1.2.3`. This caused valid builds to be incorrectly rejected.
- Makes the step solver consistent with resolvo: a namespaced var is scoped to its package, so only an exactly-matching option is relevant; base-name matching now applies **only** to un-namespaced (global) requests.
- Adds a test (`test_build_var_name_collides_with_package_name`) run against all three solvers (cli, checks, resolvo); the consumer build now succeeds for all of them.

## Root cause

`crates/spk-solve/crates/validation/src/validators/var_requirements.rs` skipped an option only when it was neither the exact request nor a base-name match:

```rust
let is_not_requested = *name != request.var;
let is_not_same_base = request.var.base_name() != name.base_name();
if is_not_requested && is_not_same_base {
    continue;
}
```

So `demo.samenameaspkg` (a namespaced var) matched the option `samenameaspkg` (a package version) because they share the base name.

## Fix

An option is relevant only when it names the exact same var, or — for un-namespaced (global) requests — when it shares the same base name. Namespaced requests must match exactly. This mirrors resolvo, which scopes namespaced var requests to their package (`is_satisfied_by(&package)`) and treats global vars as distinct pseudo-packages.

## Test plan

- [x] `cargo test -p spk-cmd-build test_build_var_name_collides_with_package_name` passes for cli, checks, and resolvo
- [x] `cargo test -p spk-cmd-build test_build_package_with_required_var_dependency` still passes (legitimate namespaced var requirement)
- [x] `cargo test -p spk-solve-validation` and `cargo test -p spk-solve var` pass (41 solver var tests, both solvers)

### DIFF
--- a/crates/spk-cli/cmd-build/src/cmd_build_test/mod.rs
+++ b/crates/spk-cli/cmd-build/src/cmd_build_test/mod.rs
@@ -859,6 +859,97 @@ build:
     .expect("Expected build of mypkg to succeed");
 }
 
+#[spfstest]
+#[rstest]
+#[case::cli("cli")]
+#[case::checks("checks")]
+#[case::resolvo("resolvo")]
+#[tokio::test]
+async fn test_build_var_name_collides_with_package_name(
+    tmpdir: tempfile::TempDir,
+    #[case] solver_to_run: &str,
+) {
+    let _rt = spfs_runtime().await;
+
+    build_package!(
+        tmpdir,
+        "demo.spk.yaml",
+        br#"
+api: v0/package
+pkg: demo/1.0.0
+
+build:
+  options:
+    - var: samenameaspkg/true
+  script:
+    - "true"
+"#,
+        solver_to_run
+    );
+
+    build_package!(
+        tmpdir,
+        "samenameaspkg.spk.yaml",
+        br#"
+api: v0/package
+pkg: samenameaspkg/1.2.3
+
+build:
+  script:
+    - "true"
+"#,
+        solver_to_run
+    );
+
+    build_package!(
+        tmpdir,
+        "middle.spk.yaml",
+        br#"
+api: v0/package
+pkg: middle/1.0.0
+
+build:
+  options:
+    - pkg: demo
+    - var: demo.samenameaspkg/true
+  script:
+    - "true"
+
+install:
+  requirements:
+    - pkg: demo
+      fromBuildEnv: true
+    - var: demo.samenameaspkg/true
+"#,
+        solver_to_run
+    );
+
+    // Building `consumer` brings both the `samenameaspkg` package and the
+    // `middle` package (which carries an install requirement for the
+    // namespaced var `demo.samenameaspkg/true`) into the same environment.
+    // A namespaced var is scoped to its package, so the `demo.samenameaspkg`
+    // var must not be confused with the unrelated `samenameaspkg` package's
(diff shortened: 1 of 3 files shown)


jrray:
@rydrman I found a case where the solver implementations disagree, the PR is a little biased to claim that resolvo is correct but this is up for debate. I could use your input because it is about the treatment of namespaced vs non-namespaced vars.

If a package _a_ defines a var `b/true` that happens to be the same name as the name of a package, a conflict arises in the event that a consumer package wants to include both packages `a` and `b` in the same environment. Resolving `b` causes the solver state to contain `b = 1.2.3` and then that prevents `a` from being resolved because `a` wants `a.b = true`.

Do we say that packages are not allowed to define vars that have the same name as other packages, which requires a certain level of omniscience, or do we start treating `b` and `a.b` as distinct vars [in the step solver]?

Resolvo already treats them as distinct and this doesn't run afoul of any of the existing solver tests. But I chalk that up to there being missing tests that demonstrate the desired behavior.

My own two cents, I think if two different packages both define a var with the same name, it shouldn't be assumed that this was intentional and that the vars are sharing a meaning or trying to coordinate. I prefer a pattern where if two or more packages want to coordinate on the same var then there should be an additional package `foo` in the mix to define the var, and then all the packages trying to coordinate all use `foo.varname` making the coordination explicit.

rydrman:
It definitely feels to me like there is a bug here, mainly in that I don't see how it's reasonable not to also consider the namespaces here when comparing the request to the var. Unless I'm not reading this code right or missing some key context about where this code is executed... Would considering the namespace as part of this resolve the conflict?

https://github.com/spkenv/spk/blob/main/crates/spk-solve/crates/validation/src/validators/var_requirements.rs#L26

jrray:
> It definitely feels to me like there is a bug here, mainly in that I don't see how it's reasonable not to also consider the namespaces here when comparing the request to the var. Unless I'm not reading this code right or missing some key context about where this code is executed... Would considering the namespace as part of this resolve the conflict?
> 
> https://github.com/spkenv/spk/blob/main/crates/spk-solve/crates/validation/src/validators/var_requirements.rs#L26

Comparing the whole thing is how resolvo is wired up which is why it passes this test. If you consider this a bug in the step solver (I prefer this stance) then I can make them both consistent by taking the resolvo behavior as correct.

jrray:
Rebased on main and addressed the review. I took @rydrman's suggestion and made the step solver consider the namespace, which makes it consistent with resolvo.

`VarRequirementsValidator` now treats a namespaced var as scoped to its package: it only matches an option of the exact same name, and base-name matching applies solely to un-namespaced (global) requests. This mirrors how resolvo wires things up — namespaced var requests become a constraint on the named package (`is_satisfied_by(&package)`), while global vars are distinct pseudo-packages (`spk_provider.rs` skips namespaced build options when populating global var values).

The test now expects success across all three solvers.

rydrman (approved):

PR #1363, opened 2026-05-08
Indexes 10: Adds index update listeners and Indexer

dcookspi:
This adds an index update listening step to spk commands that update packages: `publish`, `deprecate`, `undeprecate`, and `remove`. After finishing their current processing, these comands will wait and listen to messages on an index updates topic to verify their changes have made it into the repository's index before finishing.

This also adds the `spk repo indexer ...` command to start an index updating process (an Indexer) that will listen for package update messages for a given repo, and kick off index updates for that repo's index as packages are changed.  The Indexer will send index status event messages to the index updates topic for other spk commands to listen to. The  `publish`, `deprecate`, `undeprecate`, and `remove` commands will use those messages to determine when the index the care about has been updated.

This also extends the spk configuration for messaging channels, indexes, and indexers to include the index status topic, naming channels and indexers, and various new timeouts and settings.

The idea is to support automatic updating of an index when packages change. A site would need an external messaging system set up. But then the site could configure a message channel and an indexer, and run an indexer as a service (e.g. for the origin repo). Then all the spk commands that modify packages in that repo, would send messages, via the message channel, to the indexer for it to update the index automatically, and the indexer would send messages back to the commands to let them know the index now contains their changes. 

This is another PR in the chain for adding indexes to spk solves:
1. #1336
2. #1337
3. #1338
4. #1339
5. #1340
6. #1344
7. #1354
8. #1355
9. #1356
10. this PR
11. #1364
12. #1368
13. #1374

### DIFF
--- a/crates/spk-cli/group2/src/cmd_remove.rs
+++ b/crates/spk-cli/group2/src/cmd_remove.rs
@@ -2,8 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/spkenv/spk
 
+use std::collections::HashMap;
 use std::io::Write;
 
+use chrono::{DateTime, Utc};
 use clap::Args;
 use colored::Colorize;
 use itertools::Itertools;
@@ -11,6 +13,7 @@ use miette::{IntoDiagnostic, Result};
 use spk_cli_common::{CommandArgs, Run, flags};
 use spk_schema::foundation::format::FormatIdent;
 use spk_schema::ident::parse_ident;
+use spk_schema::name::RepositoryNameBuf;
 use spk_schema::spec_ops::WithVersion;
 use spk_schema::{BuildIdent, VersionIdent};
 use spk_storage as storage;
@@ -43,6 +46,9 @@ impl Run for Remove {
             );
             return Ok(1);
         }
+
+        let mut updated_repos: HashMap<RepositoryNameBuf, storage::RepositoryHandle> =
+            HashMap::new();
 
         for name in &self.packages {
             if !name.contains('/') && !self.yes {
@@ -84,12 +90,23 @@ impl Run for Remove {
                             remove_all(repo_name, repo, &version).await?;
                         }
                         (version, Some(build)) => {
-                            remove_build(repo_name, repo, &version.into_build_ident(build)).await?;
+                            let ident = version.into_build_ident(build);
+                            remove_build(repo_name, repo, &ident).await?;
+
+                            updated_repos.insert(repo.name().into(), repo.clone());
                         }
                     }
                 }
             }
         }
+
+        // Wait for the index to be updated before finishing
+        let utc_now: DateTime<Utc> = Utc::now();
+        let remove_time = utc_now.timestamp();
+        for (_repo_name, repo) in updated_repos.iter() {
+            repo.wait_for_index_to_update(remove_time).await?;
+        }
+
         Ok(0)
     }
 }
--- a/crates/spk-cli/group2/src/cmd_remove.rs
+++ b/crates/spk-cli/group2/src/cmd_remove.rs
@@ -5,7 +5,7 @@
 use std::collections::HashMap;
 use std::io::Write;
 
-use chrono::{DateTime, Utc};
+use chrono::Utc;
 use clap::Args;
 use colored::Colorize;
 use itertools::Itertools;
@@ -88,6 +88,8 @@ impl Run for Remove {
                     match version.into_inner() {
                         (version, None) => {
                             remove_all(repo_name, repo, &version).await?;
+
+                            updated_repos.insert(repo.name().into(), repo.clone());
                         }
                         (version, Some(build)) => {
                             let ident = version.into_build_ident(build);
@@ -101,10 +103,9 @@ impl Run for Remove {
         }
 
         // Wait for the index to be updated before finishing
-        let utc_now: DateTime<Utc> = Utc::now();
-        let remove_time = utc_now.timestamp();
+        let remove_time = Utc::now();
         for (_repo_name, repo) in updated_repos.iter() {
-            repo.wait_for_index_to_update(remove_time).await?;
+            repo.wait_for_index_to_update(&remove_time).await?;
         }
 
         Ok(0)
--- a/crates/spk-config/src/config.rs
+++ b/crates/spk-config/src/config.rs
@@ -265,9 +265,9 @@ fn default_kafka_index_update_listener_recent_past_duration_ms() -> i64 {
 
 /// Helper for a default kafka index update listener's broker fetch
 /// timeout in seconds, when not specified in config file.
-fn default_kafka_index_update_listener_broker_fetch_timeout_s() -> u64 {
+fn default_kafka_index_update_listener_broker_fetch_timeout_ms() -> u64 {
     // 20 seconds
-    20
+    20 * 1000
 }
 
 /// Configuration for using Kafka as a message channel.
@@ -322,9 +322,9 @@ pub struct KafkaChannel {
     pub index_update_listener_recent_past_duration_ms: i64,
 
     /// The data fetching timeout used by an index update listener
-    /// when querying a kafka broker, in seconds.
-    #[serde(default = "default_kafka_index_update_listener_broker_fetch_timeout_s")]
-    pub index_update_listener_broker_fetch_timeout_s: u64,
+    /// when querying a kafka broker, in milliseconds.
+    #[serde(default = "default_kafka_index_update_listener_broker_fetch_timeout_ms")]
+    pub index_update_listener_broker_fetch_timeout_ms: u64,
 }
 
 /// Types of message channels.
--- a/crates/spk-config/src/config.rs
+++ b/crates/spk-config/src/config.rs
@@ -317,7 +317,7 @@ pub struct KafkaChannel {
 
     /// How far back in time counts as recent for an index update
     /// listener when it is trying to workout if an indexer is running
-    /// for the repository index it wants to know about, in seconds.
+    /// for the repository index it wants to know about, in milliseconds.
     #[serde(default = "default_kafka_index_update_listener_recent_past_duration_ms")]
     pub index_update_listener_recent_past_duration_ms: i64,
 
--- a/crates/spk-config/src/config.rs
+++ b/crates/spk-config/src/config.rs
@@ -120,6 +120,14 @@ pub struct Index {
     /// Maximum number of times to try to get a write lock on the
     /// index data, for index generation and updates.
     pub lock_max_tries: u64,
+
+    /// Name of the configured messaging channel (see MessageChannel)
+    /// to send index status update messages too.
+    pub update_message_channel: String,
+
+    /// How often index generation and update processing should send
+    /// index update event messages.
+    pub update_event_send_freq_ms: u64,
 }
 
 impl Default for Index {
@@ -135,6 +143,9 @@ impl Default for Index {
             // for index writing.
             lock_sleep_seconds: 6,
             lock_max_tries: 5,
+            update_message_channel: String::from("kafka"),
+            // 5 seconds in milliseconds
+            update_event_send_freq_ms: 5000,
         }
     }
 }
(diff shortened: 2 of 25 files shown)


crates/spk-storage/src/storage/repository.rs
```diff
@@ -474,6 +480,13 @@ pub trait Repository: Storage + Sync {
             let new_embedded_providers = self.get_embedded_providers(package)?;
             // No change case #1: no embedded packages involved.
             if original_embedded_providers.is_empty() && new_embedded_providers.is_empty() {
+                announce_package_event(
```
dcookspi: I've realised the additional `announce_package_event` calls in this method (`update_publish`) are not correct. They all need to be closer to the points where the methods that update the storage are called, so I'll be moving these around abit.

- [x] Fix this.

dcookspi: I've moved these calls around to be closer to the underlying storage updating calls.

crates/spk-cli/group2/src/cmd_remove.rs
```diff
@@ -84,12 +90,23 @@ impl Run for Remove {
                             remove_all(repo_name, repo, &version).await?;
```
jrray, line 90: This is a material change to a repo that should be waited on too.

crates/spk-config/src/config.rs
```diff
@@ -204,22 +215,116 @@ pub struct HostOptions {
...
+    /// How far back in time counts as recent for an index update
+    /// listener when it is trying to workout if an indexer is running
+    /// for the repository index it wants to know about, in seconds.
+    #[serde(default = "default_kafka_index_update_listener_recent_past_duration_ms")]
+    pub index_update_listener_recent_past_duration_ms: i64,
+
+    /// The data fetching timeout used by an index update listener
+    /// when querying a kafka broker, in seconds.
+    #[serde(default = "default_kafka_index_update_listener_broker_fetch_timeout_s")]
+    pub index_update_listener_broker_fetch_timeout_s: u64,
```
jrray: I think it would be less accident prone to keep all these properties consistent and use ms units here too.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -2,39 +2,135 @@
...
+    PackageEventMessage,
+};
+use crate::{Error, FlatBufferRepoIndex, RepositoryHandle, RepositoryIndexMut, Result};
 
-/// The queue timeout for a producer sending a message to a queue.
-/// This determines how to long retry for if the producer queue is full.
-const PRODUCER_QUEUE_TIMEOUT: Duration = Duration::from_secs(4);
+/// Number of milliseconds to sleep while waiting for next message,
+/// when listening for new messages.
+const LISTEN_YIELD_SLEEP_TIME: u64 = 100;
```
jrray: Outside of config properties I would really prefer using the `Duration` type for storing durations.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -2,39 +2,135 @@
...
+
+/// Number of milliseconds in a second, used in time conversions
+const NUM_MS_IN_ONE_SECOND: u64 = 1000;
```
jrray: I predict using `Duration` will make this const unnecessary.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -57,25 +153,560 @@ pub(crate) async fn announce_package_event(
...
+}
+
+/// Listen to the index updates topic until there's an "index completed"
+/// message for the given repository's index with an index update
+/// start time after the given update timestamp. This will also
+/// timeout if too much time passes without receiving any index update
+/// messages.
+pub(crate) async fn listen_to_index_status_updates(
+    kafka_channel: &KafkaChannel,
+    min_update_time: i64,
```
jrray: `DateTime<Utc>` would be a good choice of type for this parameter.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -57,25 +153,560 @@ pub(crate) async fn announce_package_event(
...
+    let mut there_is_a_recent_message = false;
+    let mut time_since_last_message = 0;
+
+    tracing::debug!(
+        "Looking for a recent index update message that was sent after: {min_update_time}"
+    );
+
+    let mut stream = consumer.stream();
+
+    while running.load(std::sync::atomic::Ordering::SeqCst) {
```
jrray: It looks like `Relaxed` would be fine for all the uses of this atomic. It's not being used to order reads or writes of other memory.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -57,25 +153,560 @@ pub(crate) async fn announce_package_event(
...
+
+    // Start listening to index update messages
+    let mut time_since_last_message = 0;
+    let mut read_messages = 0;
+    let mut generate_full_index = false;
+    let mut package_versions: HashSet<OptVersionIdent> = HashSet::new();
+
+    let mut stream = consumer.stream();
+
+    while running.load(std::sync::atomic::Ordering::SeqCst) {
```
jrray: Same thoughts as above about `Relaxed`.

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -57,25 +153,560 @@ pub(crate) async fn announce_package_event(
...
+        tokio::select! {
+            maybe_message = stream.next() => {
+                match maybe_message {
+                    Some(Ok(message)) => {
+                        // Get the package event details
+                        let Some(payload) = message.payload_view::<str>().map(|s| s.map(|s| s.to_owned()).expect("payload is legal UTF-8")) else { continue; };
```
jrray: Reading a message out of kafka is reading untrusted inputs, I don't like the idea of having an `expect` here making assumptions. You can skip messages that aren't valid instead.

crates/spk-storage/src/storage/messaging/mod.rs
```diff
@@ -4,35 +4,256 @@
...
+    /// message first started.
+    pub start: i64,
+}
+
+/// Send an index event to each configured messaging system
+pub(crate) async fn announce_index_event(
+    event: IndexEvent,
+    to: &url::Url,
+    repo_name: &RepositoryName,
+    index_start_time: i64,
```
jrray: `DateTime<Utc>`?

crates/spk-storage/src/storage/messaging/mod.rs
```diff
@@ -4,35 +4,256 @@
...
+
+    Ok(())
+}
+
+/// Subscribes to index updates topic and listens to updates until it
+/// gets one that indicates the repo's index has been updated after
+/// the given update time.
+pub(crate) async fn listen_to_index_status_until_updated(
+    repo: &RepositoryHandle,
+    update_time: i64,
```
jrray: `DateTime<Utc>`?

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -362,6 +351,41 @@ impl FileLock {
...
+                Ok(metadata) => match metadata.modified() {
+                    Ok(time) => {
+                        // Turn it into seconds since epoch timestamp
+                        let duration = time
+                            .duration_since(std::time::UNIX_EPOCH)
+                            .expect("Lock file time should be after the start of the epoch");
+                        Ok(duration
+                            .as_secs()
+                            .try_into()
+                            .expect("timestamp since epoch should be convertible"))
```
jrray: Boo! Return the richer datatype.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -584,6 +616,8 @@ impl FlatBufferRepoIndex {
     /// Gather packages and global vars from the given repos.
     async fn gather_all_data_from_repo(
         repos: &Vec<(String, crate::RepositoryHandle)>,
+        send_index_event_period_ms: Option<u64>,
+        index_start_time: i64,
```
jrray: I'm going to stop commenting on every one of these...

crates/spk-storage/src/fixtures.rs
```diff
@@ -96,10 +96,22 @@ pub async fn spfsrepo() -> TempRepo {
     make_repo(RepoKind::Spfs).await
 }
 
+pub fn disable_messaging_channels_for_tests() {
```
jrray, line 99: Hmm, this is unfortunate because technically now every test that ends up calling this should have the `#[serial_test::serial(config)]` attribute on it. This is used to prevent tests stomping on each other's attempt to change the config.

jrray (changes requested):

crates/spk-cli/group2/src/cmd_remove.rs
dcookspi: I've added this.

crates/spk-config/src/config.rs
dcookspi: I've switched the units on this value.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've changed this over.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated these.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/messaging/mod.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/messaging/mod.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: It's used on a raw config value to get a rough seconds value (line 427-ish). Ideally, it would do a milliseconds number to human readable time conversion so the log message was better.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated these places to check for the error, send it on to sentry, and discard the message.

crates/spk-storage/src/fixtures.rs
dcookspi: After our discussion, I've put in a ticket for the larger question of config and test isolation that mentions this case, and the existing metadata case it was based on, see https://github.com/spkenv/spk/issues/1379

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I'm going to look for a crate that helps with this.

dcookspi: I've updated this to use the `millisecond` crate, which has a pretty print output that doesn't round to the largest unit. I tried out `millis`, `time-format`, and `ms-convert`, but wasn't happy with their results and rounding.

crates/spk-config/src/config.rs
```diff
@@ -204,22 +215,116 @@ pub struct HostOptions {
...
+    #[serde(default = "default_kafka_index_update_listener_session_timeout_ms")]
+    pub index_update_listener_session_timeout_ms: u64,
+
+    /// The maximum polling interval for an index update listener in milliseconds
+    #[serde(default = "default_kafka_index_update_listener_max_polling_interval_ms")]
+    pub index_update_listener_max_polling_interval_ms: u64,
+
+    /// How far back in time counts as recent for an index update
+    /// listener when it is trying to workout if an indexer is running
+    /// for the repository index it wants to know about, in seconds.
```
```suggestion
    /// for the repository index it wants to know about, in milliseconds.
```

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -2,39 +2,133 @@
...
+    IndexUpdateMessage,
+    PackageEvent,
+    PackageEventMessage,
+};
+use crate::{Error, FlatBufferRepoIndex, RepositoryHandle, RepositoryIndexMut, Result};
 
-/// The queue timeout for a producer sending a message to a queue.
-/// This determines how to long retry for if the producer queue is full.
-const PRODUCER_QUEUE_TIMEOUT: Duration = Duration::from_secs(4);
+/// Number of milliseconds to sleep while waiting for next message,
```
```suggestion
/// Amount of time to sleep while waiting for next message,
```

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -57,25 +151,582 @@ pub(crate) async fn announce_package_event(
...
+        repo_to_index.address(),
+        indexer_name,
+        // This does not have an index start time so
+        // this is a placeholder.
+        &Utc::now(),
+    )
+    .await?;
+
+    // Start listening to index update messages
+    let mut time_since_last_message = 0;
```
jrray: Making this `Duration` would simplify some of this code, `Duration`s can be added and compared. Optional, since I've abused you on this topic enough.

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -584,6 +608,8 @@ impl FlatBufferRepoIndex {
     /// Gather packages and global vars from the given repos.
     async fn gather_all_data_from_repo(
         repos: &Vec<(String, crate::RepositoryHandle)>,
+        send_index_event_period_ms: Option<u64>,
```
jrray: Wishing this used `Duration`...

crates/spk-storage/src/storage/flatbuffer_index.rs
```diff
@@ -697,6 +763,8 @@ impl FlatBufferRepoIndex {
         &self,
         repo: &crate::RepositoryHandle,
         package_versions: &[OptVersionIdent],
+        send_index_event_period: u64,
```
jrray: `Duration` some more? This var name doesn't encode the expected units so I have an elevated concern about this one.

jrray (changes requested):
Thanks for making all those changes, just a few more to nag about here.

crates/spk-config/src/config.rs

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated both the functions that had these variables.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

crates/spk-storage/src/storage/flatbuffer_index.rs
dcookspi: I've updated this.

jrray (approved):

PR #1364, opened 2026-05-14
Indexes 11:  Adds sentry support for kafka message issues

dcookspi:
This adds sentry support so the kafka messaging can send data to sentry when a consumer cannot parse, or get valid data from, the message it got.

This is an indexes related PR that relies on the these indexes PRs and having the `sentry` feature enabled:

9. #1356
10. #1363
11. This PR.
12. #1368
13. #1374


dcookspi:
Todo:

- [x] Move the data in the logging breadcrumbs into extra data blocks on the sentry scope so it appears in a more useful way in sentry.

dcookspi:
Updated the breadcrumbs data.

jrray (approved):

PR #1368, opened 2026-05-26
Indexes 12: implements get_build_requirements() for IndexedPackage

dcookspi:
This implements the `get_build_requirements()` method for the `IndexedPackage` type using existing data in the index. The method will no longer error as unimplemented.

This is an indexes related PR, previous indexes PRs include:

9. #1356
10. #1363
11. #1364
12. This PR.
13. #1374

### DIFF
--- a/crates/spk-schema/src/v0/indexed_package.rs
+++ b/crates/spk-schema/src/v0/indexed_package.rs
@@ -16,7 +16,6 @@ use spk_schema_foundation::ident::{
     PinnedValue,
     PkgRequestWithOptions,
     RequestWithOptions,
-    RequestedBy,
 };
 use spk_schema_foundation::ident_build::Build;
 use spk_schema_foundation::ident_component::Component;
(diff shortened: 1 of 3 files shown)


crates/spk-schema/src/v0/indexed_package.rs
```diff
@@ -496,10 +497,35 @@ impl Package for IndexedPackage {
     }
 
     fn get_build_requirements(&self) -> crate::Result<Cow<'_, RequirementsList<PinnedRequest>>> {
-        Err(Error::SpkIndexedPackageDoesNotImplement(
-            "Package".to_string(),
-            "get_build_requirements".to_string(),
-        ))
+        let mut requests = RequirementsList::default();
+        // A duplicated of get_build_requirements() from
+        // v0/package_spec with this line changed to call a method.
```
jrray: We can avoid drift by moving this logic into a free function shared by both places.

jrray (changes requested):

crates/spk-schema/src/v0/indexed_package.rs
dcookspi: I've created a function in the v0/package_spec file and used it in both places.

jrray (approved):
Perfect!

PR #1372, opened 2026-05-27
Fix spfstest reexec race

jrray:
This PR isolates the `#[spfstest]` wrapper fix that stopped one failing inner test from poisoning later sibling cases. Keeping the re-exec coordination change on its own makes the failure-mode discussion small and reviewable before the solver and `v1/platform` work stack on top of it.

## Summary
- serialize the outer `#[spfstest]` re-exec handoff across sibling tests
- propagate the shared re-exec exit status so an inner failure cannot leave the outer test process in a successful state
- add regression coverage that asserts failing inner tests still fail the wrapper process while unrelated sibling bodies still run, without depending on interleaved `--nocapture` libtest output, plus unit coverage for exit-status propagation

## Testing
- cargo test -p spfstest multiple_spfstest_cases_exit_nonzero_when_inner_tests_fail -- --nocapture (run repeatedly)
- cargo test -p spfstest -- --nocapture
- spk env nodejs -- node "$HOME/.local/share/nvim/mason/bin/cspell" lint --config cspell.json crates/spfstest/src/lib.rs


jrray:
@dcookspi I added a test that exercises when multiple test cases share the same process / spfs runtime that all the cases execute and the success an failure counts are as expected.

jrray:
The new test also tries to prove the test process won't exit with 0 if any tests fail.

dcookspi (approved):

PR #1373, opened 2026-05-27
Fix build-component requirement handling for resolved and embedded packages

jrray:
## Summary

Build-component requirements (for example an `IfAlreadyPresent` constraint declared on a package's `build` component) were not represented in the build context. As a result a later requirement could tighten the merged request for an **already-resolved** package without the solver noticing, and the resolved package would silently violate the tightened request. This PR makes both solver flavors honor those constraints, including when the constrained package is provided as an embedded build.

## Step solver

- Keep build-component requests in the build context.
- Validate each decision **before it is committed** (in `reconsider_invalidated_resolve`) instead of re-proving every resolved package on every step, so states stay valid by construction. When a decision would tighten the request for an already-resolved package such that its resolved build no longer satisfies the version:
  - if the package is a plain repository build whose request merely needs tightening, step back to before it was resolved and re-request it with the tightened constraint (mirroring the existing `ConflictingEmbeddedPackage` handling);
  - otherwise abandon the branch and let the solver backtrack — this covers both the **dependent-package** case (a different resolved package pinned a conflicting version) and the **embedded** case (the stale package is provided by, and so governed by, its parent).

## Resolvo solver

- An `IfAlreadyPresent` requirement constrained the specific (build) component package, which misses a package that is only present under a different component — such as one provided as an embedded build with no separate build component. Constrain the synthetic `Base` component instead, so the version binds whenever the package is in the solution regardless of which component pulled it in.

## Tests

New regressions across both solver flavors (and both index modes):

- `..._reconsiders_existing_build_request` — a broad build request tightened by a later build-component `IfAlreadyPresent` constraint.
- `..._reconsiders_dependent_package` — a dependent package that must be reconsidered when the dependency is re-pinned.
- `..._reconsiders_embedded_package` — an embedded package whose **provider** must be reconsidered to satisfy the tightened constraint.

## Testing

- `cargo test -p spk-solve`
- `cargo clippy -p spk-solve --all-targets`
- `make format`

### DIFF
--- a/crates/spk-solve/src/solvers/step/solver.rs
+++ b/crates/spk-solve/src/solvers/step/solver.rs
@@ -934,6 +934,13 @@ impl Solver {
                         }
                     };
 
+                    if let Some(reconsider) = self
+                        .reconsider_invalidated_resolve(graph, &node.state, &decision)
+                        .await?
+                    {
+                        decision = reconsider;
+                    }
+
                     decision.add_notes(notes.iter().cloned());
                     return Ok(Some(decision));
                 }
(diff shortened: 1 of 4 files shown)


crates/spk-solve/src/solvers/step/solver.rs
```diff
@@ -475,6 +481,7 @@ impl Solver {
         node: &mut Arc<Node>,
     ) -> Result<Option<Decision>> {
         let mut notes = Vec::<Note>::new();
+        self.ensure_resolved_packages_are_valid(&node.state)?;
```
jrray: I think this is weird. It implies that it is possible to create a new state that isn't already proved valid.

dcookspi (approved):

dcookspi (approved):

PR #1374, opened 2026-05-27
Indexes 13: Bugfix for missing package sizes in solution summary from an index

dcookspi:
This fixes a bug introduced by using an index that meant all the packages would be shown as zero size in the solution summary output. This restores the package component disk usage sizes by exposing the underlying repo in the index, and using that when calculating sizes at the end of a solve.

This is an indexes related PR, previous indexes PRs include:

9. #1356
10. #1363
11. #1364
12. #1368
13. This PR.

### DIFF
--- a/crates/spk-storage/src/storage/indexed.rs
+++ b/crates/spk-storage/src/storage/indexed.rs
@@ -18,6 +18,7 @@ use spk_schema::ident_build::EmbeddedSource;
 use spk_schema::name::OptNameBuf;
 use spk_schema::{BuildIdent, Spec, SpecRecipe};
 
+use super::RepositoryHandle;
 use super::repository::{PublishPolicy, Repository, Storage};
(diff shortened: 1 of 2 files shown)


rydrman (approved):

crates/spk-storage/src/storage/indexed.rs
```diff
@@ -68,6 +69,11 @@ impl IndexedRepository {
         self.wrapped_repo.index_location_path().await
     }
 
+    /// Get the underlying repo handle this index wrapper.
```
jrray:
```suggestion
    /// Get the underlying repo handle this index wraps.
```

jrray (changes requested):

crates/spk-storage/src/storage/indexed.rs
dcookspi: I've updated this.

jrray (approved):

PR #1383, opened 2026-06-04
Changes use index default in the config to false

dcookspi:
This changes the `use_index` default  config for repositories to false. So indexes won't be used by default.


jrray:
We decided it would be safer to not enable by default, to allow for organizations to be able to safely work on setting up the indexing system and verify it is updating properly before end users start to consume the index.

We plan to take this approach and enable indexes via the spk config file once we are happy with our indexing pipeline.

jrray (approved):

PR #1384, opened 2026-06-05
Add kmod dependency and document running spfs in containers

jrray:
## Summary

spfs shells out to `/sbin/modinfo overlay` at runtime to detect which `overlayfs` mount options the kernel supports. In minimal container images `modinfo` (provided by `kmod`) is absent, so spfs falls back to a conservative set of mount options and logs a warning on every run:

```
WARN Failed to detect supported overlayfs params: Failed to spawn /sbin/modinfo
WARN  > Falling back to the most conservative set, which is undesirable
```

This PR addresses that two ways:

- **rpm specs** — declare `kmod` as a runtime `Requires` in `spfs.spec`, `spk.spec`, and `.site/spi/spk.spec`, so rpm-based installs always provide `/sbin/modinfo`. (The specs previously listed `fuse3`/`rsync` but not `kmod`, despite spfs calling `modinfo` at runtime.)
- **docs** — add `docs/spfs/container.md`, a guide for running spfs inside **Docker** and **Podman**, linked from the spfs section index.

## The container fix (documented in the guide)

Avoiding the warning without `SPFS_SUPPRESS_OVERLAYFS_PARAMS_WARNING` requires two things, both covered in the guide:

1. Install `kmod` in the image (provides `/sbin/modinfo`).
2. Bind-mount the host's modules read-only (`-v /lib/modules:/lib/modules:ro`) so `modinfo overlay` finds metadata for the running (host) kernel. This is version-agnostic, unlike baking a fixed `kernel-modules` package into the image.

The guide also documents the `--privileged` / `CAP_SYS_ADMIN` requirement for the overlay mount, and that **rootless Podman** cannot mount `/spfs` even when privileged (use rootful Podman or Docker).

## Testing

Verified end-to-end with spfs built natively in a Rocky Linux 9.6 container under Podman: the stock image shows the warning; with `kmod` installed and `/lib/modules` bind-mounted, `modinfo overlay` exits 0 and `spfs run` completes with no warning.


rydrman (approved):
Love it!

PR #1387, opened 2026-06-08
spk ls Adds repo name to 'spk ls pkgname -v'.

dcookspi:
This makes 2 changes to `spk ls pkgname ...` output:
- It adds the repo names to the version number list when `-v` is used
- It depduplicates and sorts the version number list when `-v` is not used

For example:
```
> spk ls pkgname -v
 [local] 0.3.1
[origin] 0.3.5
[origin] 0.3.2
[origin] 0.3.1
```
and 
```
> spk ls pkgname
0.3.5
0.3.2
0.3.1
```

### DIFF
--- a/crates/spk-cli/group2/src/cmd_ls.rs
+++ b/crates/spk-cli/group2/src/cmd_ls.rs
@@ -377,7 +377,7 @@ impl<T: Output> Ls<T> {
                 .map(|(prefix, version, label)| format!("{prefix}{version}{label}"))
                 .collect())
         } else {
-            // If -v is not used, sort deduplicate and the versions
+            // If -v is not used, sort and deduplicate the versions
             // from multiple repos together, and discard their
             // prefixes (repo names).
             Ok(lines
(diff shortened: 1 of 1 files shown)


crates/spk-cli/group2/src/cmd_ls.rs
```diff
@@ -307,49 +322,87 @@ impl<T: Output> Ls<T> {
...
+
+                    lines.push(line_data);
                 }
                 _ => {}
             }
         }
 
-        Ok(lines)
+        if self.verbose > 0 {
+            // Include the repo name and do no deduplicate or re-sort
```
jrray:
```suggestion
            // Include the repo name and do not deduplicate or re-sort
```

crates/spk-cli/group2/src/cmd_ls.rs
```diff
@@ -307,49 +322,87 @@ impl<T: Output> Ls<T> {
...
+            // the versions.
+            Ok(lines
+                .into_iter()
+                .map(|(prefix, version, label)| format!("{prefix}{version}{label}"))
+                .collect())
+        } else {
+            // If -v is not used, sort deduplicate and the versions
```
jrray:
```suggestion
            // If -v is not used, sort and deduplicate the versions
```

jrray (approved):

PR #1389, opened 2026-06-19
Add two initial metrics to the indexer

dcookspi:
This adds some initial metrics to the indexer. It moves the metrics code to the `spk-config` crate to do this, and avoid creating a dependency cycle between the spk-solver and spk-storage crates.

This adds two metrics:
- one after each indexer heartbeat message is sent
- one just before starting any index update


crates/spk-config/src/metrics.rs
```diff
@@ -103,6 +103,18 @@ pub static SPK_SOLVER_SOLUTION_SIZE_METRIC: Lazy<String> = Lazy::new(|| {
...
+// TODO: add the default value to a config file, once spk has one
+pub static SPK_INDEXER_HEARTBEAT_METRIC: Lazy<String> = Lazy::new(|| {
+    std::env::var("SPK_INDEXER_HEARTBEAT_METRIC")
+        .unwrap_or_else(|_| String::from("spk.indexer_heartbeat_count"))
+});
+
+// TODO: add the default value to a config file, once spk has one
+pub static SPK_INDEXER_INDEX_UPDATE_METRIC: Lazy<String> = Lazy::new(|| {
+    std::env::var("SPK_INDEXER_INDEX_UPDATE_METRIC")
+        .unwrap_or_else(|_| String::from("spk.indexer_index_update_count"))
```
jrray: Instead of having hard-coded fallbacks, please make it so sending these count changes is disabled unless the metric names are found in the configuration.

jrray (changes requested):

crates/spk-config/src/metrics.rs
dcookspi: - [x] Remove the indexer metric names, change to not send indexer metrics unless the names are configured.

dcookspi: I've removed the default metrics names and moved them to the Indexer config section. The metrics won't be sent unless the metric names are configured. Getting the correct metric name to the point where the event messages are sent required adding the `IndexUpdateContext` helper struct.

crates/spk-config/src/config.rs
```diff
@@ -375,6 +375,21 @@ pub struct Indexer {
     /// Indexer's kafka broker maximum polling interval in milliseconds
     #[serde(default = "default_indexer_max_polling_interval_ms")]
     pub max_polling_interval_ms: u64,
+
+    /// Name of the metric to update whenever the indexer sends an
+    /// index event message (including heartbeats). With suitable
+    /// external systems, this can be used to monitor an indexer's
+    /// workload and whether it is healthy. If it is not set the, or
+    /// statsd is not configured, metric will not be updated.
```
```suggestion
    /// workload and whether it is healthy. If it is not set, or
    /// statsd is not configured, the metric will not be updated.
```

crates/spk-storage/src/storage/messaging/kafka.rs
```diff
@@ -120,7 +123,8 @@ pub(crate) async fn announce_index_event(
     event: IndexEvent,
     to: &url::Url,
     repo_name: &RepositoryName,
-    index_start_time: &DateTime<Utc>,
+    index_context: &IndexUpdateContext,
+    //index_start_time: &DateTime<Utc>,
```
```suggestion
    index_context: &IndexUpdateContext,
```

jrray (changes requested):

crates/spk-config/src/config.rs

crates/spk-storage/src/storage/messaging/kafka.rs
dcookspi: I've updated this.

PR #1391, opened 2026-06-24
Fix intermittent IllegalGeneration error in spk publish

jrray:
## Problem

`spk publish` sometimes fails with:

```
ERROR   × failed to commit new starting offsets for the kafka index updates
  │ consumer: Consumer commit error: IllegalGeneration (Broker: Specified
  │ group generation id is not valid)
```

## Root cause

`listen_to_index_status_updates()` (run by `spk publish` to wait for the index to update) `subscribe()`d to the index-updates topic via Kafka **consumer-group management**, then immediately `commit()`ed the starting offsets — *before* the consumer had joined the group / finished rebalancing. The group join only happens on the first poll (`stream.next()`), so the commit was sent with no valid group generation, yielding `IllegalGeneration`. It's intermittent because it's a timing race against the join/rebalance.

The commit also served no purpose: this consumer uses a fresh, ephemeral `Ulid` group id every run, so the committed offsets are never read back.

## Fix

Switch this read-only, ephemeral consumer from `subscribe()` + `commit()` to manual partition **`assign()`**, carrying the starting offsets in the `TopicPartitionList`. Manual assignment has no group generation, so the race cannot occur. The `subscribe()` call is removed (a consumer can't mix `subscribe` and `assign`).

The indexer's long-lived consumer (`listen_to_package_events_and_run_index_updates`) is untouched — it legitimately uses a stable group id and offset commits.

## Testing

Verified by `cargo check -p spk-storage`. Not reproduced end-to-end against a live Kafka broker.


dcookspi:
Related to https://github.com/spkenv/spk/pull/1392 . I rebased this onto https://github.com/spkenv/spk/pull/1392  but cannot edit the PR to change it from main, so the diffs include both sets of changes now. I won't matter as long as both are approved. But this is a note to remember this when it comes to merging.

Makefile
```diff
@@ -90,6 +90,12 @@ nextest: FEATURES?=server,spfs/server
 nextest:
 	cargo nextest run $(cargo_features_arg) $(cargo_packages_arg) $(TEST_ARGS)
 
+# Run the spk-storage kafka integration tests against a broker started
```
jrray: This is still pending removal, and some of the words added to the dictionary are likely not necessary with removing the live kafka test.

dcookspi: I've removed this, an extrra dependency, and the unused word from the dictionary.

jrray:
I can't approve this because I submitted it.

dcookspi (approved):

